PR #11, opened 2018-04-26
Search-813 Update ACS 6.0 to use ASS 1.1.1 and Remove References to Solr4

meenalbhave:
Update the docker packaging to use ASS version 1.1.1

Remove the references to older solr versions that will not be
 supported anymore with ACS 6.0.

### DIFF
--- a/distribution/src/assembly/distribution.xml
+++ b/distribution/src/assembly/distribution.xml
@@ -33,22 +33,6 @@
             </includes>
         </fileSet>
 
-        <!-- Solr4 Distribution -->
-        <fileSet>
-            <directory>${project.build.directory}/dependency/solr4</directory>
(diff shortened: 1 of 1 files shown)


derekhulley:
This looks good.
The test docker compose is updated and SOLR4 is completely removed from the packaging.  There is no need, just for clarity, for this packaging project to distribute Search Services.
LGTM.

michaelsuzukisagi:
@meenalbhave did you verify that when we run the alfresco from the distribution its defaulted to search services(solr6).

meenalbhave:
I have verified that alfresco distribution does not contain any solr / search artifacts. Search Services settings in alfresco-global.properties file will dictate the version of Solr / Search used by ACS.

PR #17, opened 2018-05-18
REPO-3275 secure tomcat

Epurashu:
Unpack root.war
   Grant alfresco all security permissions
   Grant root.war only permission to deploy

### DIFF
--- a/docker-alfresco/Dockerfile
+++ b/docker-alfresco/Dockerfile
@@ -42,6 +42,13 @@ RUN java -jar /usr/local/tomcat/alfresco-mmt/alfresco-mmt*.jar install \
 
 # Docker CMD from parent image starts the server
 
+# Unpack ROOT.war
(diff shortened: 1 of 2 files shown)


docker-alfresco/Dockerfile
```diff
@@ -42,6 +42,13 @@ RUN java -jar /usr/local/tomcat/alfresco-mmt/alfresco-mmt*.jar install \
 
 # Docker CMD from parent image starts the server
 
+# Unpack ROOT.war
+RUN mkdir /usr/local/tomcat/webapps/ROOT && cd /usr/local/tomcat/webapps/ROOT && jar -xvf /usr/local/tomcat/webapps/ROOT.war && rm -f /usr/local/tomcat/webapps/ROOT.war && \
```
derekhulley: Is it possible to make the directory read-only afterwards?

Epurashu: Done.

PR #18, opened 2018-05-22
GOOGLEDOCS-429 : Generate RC2 for GoogleDocs 3.1.0 compatible with ACS Ent 6.0

DenisGabriela:
- updated googledocs dependencies


docker-alfresco/pom.xml
```diff
@@ -153,11 +152,18 @@
...
                                     <artifactId>alfresco-googledocs-repo</artifactId>
                                     <version>${alfresco.googledocs.version}</version>
-                                    <classifier>enterprise</classifier>
                                     <type>amp</type>
                                     <overWrite>false</overWrite>
                                     <outputDirectory>${project.build.directory}/amps</outputDirectory>
                                 </artifactItem>
+                                <artifactItem>
+                                    <groupId>org.alfresco.integrations</groupId>
+                                    <artifactId>alfresco-googledocs-share</artifactId>
```
andrei-rebegea: share amp needs to be applied to the share docker image : https://github.com/Alfresco/share/blob/6.0/packaging/docker/pom.xml#L33

this image is for alfresco-content-repository - backend only

DenisGabriela: Thanks. I removed it and I will open a pull request on share as well.

PR #36, opened 2018-07-05
REPO-3590: 6.0: Include Document Transformation Engine amp into "all amps" test

aforascu:
- fixed share on all amps test image (share was not starting)
- update DTS version because 2.2.0 is not available on nexus


andrei-rebegea:
looks good

PR #41, opened 2018-07-27
REPO-3696 Add Shared File Store to deployment

alandavis:


alandavis:
Having spoken to Alex, a further commit was made to use the latest version of transformers and SFS when run from BAT but a known version in Bamboo.

PR #50, opened 2018-09-24
Upgrade Spring 5.1.0 and webscripts to 6.20

derekhulley:


derekhulley:
Changes need to be picked over to SP/6.0.N as well.

PR #53, opened 2018-09-28
Feature/repo 3829 aws image

mmuller88:
* Create aws image with s3 amp and mariadb driver in it


mmuller88:
Spoke with Alex. We shouldn't use two different Dockerfiles with the same content, because there can become stale . I will try to find a way where we do not need that. Unlucky simply referring to the existing Dockerfile from docker-alfresco doesn't work because it takes the docker-alfresco/target folder but we need the docker-alfresco-aws/target folder.

mmuller88:
I created a second approach where we do not need to duplicate the Dockerfile 

https://github.com/Alfresco/acs-packaging/pull/54

We need to decide which approach we take.

PR #54, opened 2018-10-01
REPO-3829 aws image option 2

mmuller88:
This is a second option as feature/REPO-3829_aws_image where the aws module is a child from docker-alfresco. With this it is possible to pick up the Dockerfile from docker-alfresco.


docker-alfresco/Dockerfile
```diff
@@ -2,6 +2,9 @@
 # More infos about this image: https://github.com/Alfresco/alfresco-docker-base-tomcat
 FROM alfresco/alfresco-base-tomcat:8.5.28-java-8-oracle-centos-7
 
+# Set default docker_context. Will / Can be overriden with maven.
+ARG docker_context=target
```
killerboot: Please change the name of the variable to `resource_path` or similar. `docker_context` can be misinterpreted as build context, see [docs](https://docs.docker.com/engine/reference/commandline/build/#extended-description).

PR #58, opened 2018-10-11
REPO-3811 Fix broker URLs

killerboot:


eknizat:
Very nice Alex! Just tried it out and repo is now connecting to active mq

PR #71, opened 2018-10-16
ATS-168 : Release ATS 0.2.1-EA for ACS 6.1.0-EA3

CezarLeahu:
- update versions


montgolfiere:
Also SFS should be 0.3 ?

Otherwise, LGTM once we have a green branch build.

CezarLeahu:
I've updated the SFS_TAG, also.

PR #79, opened 2018-10-23
3rdparty updates

derekhulley:


derekhulley:
3rd party updates.
Green (apart from tests that should have been quarantined).  Any reason NOT to merge?

PR #81, opened 2018-10-26
REPO-3944  Renditions: Switches for TransformServer and Local Transforms

alandavis:
Doc changes


alandavis:
Also includes the upgrade to SFS 0.4 which appears to have been missed in other work.

alandavis:
Derek is okay with this

PR #100, opened 2018-11-13
Add Hazelcast to dependency management

killerboot:


killerboot:
It is a good point. I can see that hazelcast doesn't have any compile dependencies:
https://mvnrepository.com/artifact/com.hazelcast/hazelcast/3.9.2

derekhulley:
Good citizens using _provided_.
LGTM.

PR #126, opened 2018-12-13
RM-6679 Integration with Glacier Connector 1.0.0-EA1

cagache:
Updated S3 connector version to 2.3.0 and added Glacier Connector dependency


aforascu:
LGTM

PR #127, opened 2018-12-13
RM-6679 Update s3 connector version

cagache:
Update S3 Connector version to 2.3.0


derekhulley:
Note that we merge to master and then cherry pick back.
In our case, we have not gone through the full test cycle with V2.3, so I suggest that you get approval for a change to the version of S3 supported before attempting to stick it into ACS 6.1.

cagache:
Thank you for looking into this. I have already pushed on master branch the new S3 connector 2.3.0 version along with Glacier Connector new dependency (1.0.0-EA1) . What is the process to get this approved to get it into the HF branch?

derekhulley:
In fact, use a cherry pick from master to the  support or HF branches, if it's OK to do.
We'll discuss the process offline.

PR #144, opened 2019-01-09
Update Share and Share Services version to RC3

alexbalmus:


aforascu:
The RC change is needed also on test env files: https://github.com/Alfresco/acs-packaging/blob/master/docker-alfresco/test/.env / https://github.com/Alfresco/acs-packaging/blob/master/docker-alfresco/test/test.env

PR #154, opened 2019-01-15
updated alfresco-repository dependency version

eliaporciani:


tunaaksoy:
Elia, you need to cherry pick your change to this branch https://github.com/Alfresco/alfresco-repository/blob/support/SP/7.33.N/pom.xml then release and use that version. By the looks of it you try to use the version from alfresco-repository master (v. 7.4x) in 6.1.N branch. Hence the jump in version from 7.3x to 7.4x. At least that's my understanding. @killerboot is this correct?

eliaporciani:
Released alfresco-repository support/SP/7.33.N version 7.33.11 with the changes of the fix/MNT-20134 and updated the alfresco-repository dependency in acs-packaging.

tunaaksoy:
LGTM. Please merge once the build is green.

PR #186, opened 2019-02-08
Add license file for YourKit

skopf:


skopf:
We are unpacking the entire YourKit redist zip file, which already contains the license file. So there is no need to add this license to our 3rd party licenses.

alandavis:
We have decided to merge in anyway as this location is where people will look and find it is not there.

PR #199, opened 2019-02-25
REPO-4217 Add missing core configuration

mmuller88:
* add missing cors.supportedHeaders range
* cors.exposedHeaders


war/src/main/webapp/WEB-INF/web.xml
```diff
@@ -160,6 +164,7 @@
       <url-pattern>/service/*</url-pattern>
       <url-pattern>/s/*</url-pattern>
       <url-pattern>/cmisbrowser/*</url-pattern>
+      <url-pattern>/definitions/*</url-pattern>
```
ancutam: What was `definitions` about in the end?

PR #200, opened 2019-02-26
[SEARCH-1272] updated repository depdendency version.

eliaporciani:
Updated alfresco-repository dependency version.


tpage-alfresco:
For clarity here - @eliaporciani and the Sease team already have a contract with Alfresco, so the CLA is not really needed in this case. However Elia - PRs will be less noisy if you can go and click to agree to the contribution agreement.

PR #216, opened 2019-03-18
Fix/repo 4202 ais docker

andrei-zapodeanu-alfresco:


docker-alfresco/ais/docker-compose.yml
```diff
@@ -0,0 +1,173 @@
...
+                -Dtika.url=http://tika:8090/
+                -Dsfs.url=http://shared-file-store:8099/
+                -Dshare.host=localhost
+                -Dmessaging.broker.url=\"failover:(tcp://activemq:61616)?timeout=3000&jms.useCompression=true\"
+                -Dlocal.transform.service.enabled=${LOCAL_TRANSFORM_SERVICE_ENABLED}
+                -Dtransform.service.enabled=${TRANSFORM_SERVICE_ENABLED}
+                -Dauthentication.chain=identity-service1:identity-service,ldap:ldap,alfrescoNtlm1:alfrescoNtlm
+                -Didentity-service.enable-basic-auth=true
+                -Didentity-service.authentication.validation.failure.silent=false
+                -Didentity-service.auth-server-url=http://<HOST_IP>:8999/auth
```
mmuller88: Do you not need to point to the created service alfresco-identity-service ?

andrei-zapodeanu-alfresco: The HOST_IP is just a placeholder. I added a comment about this:

# The identity service address should be the same as the origin of the request ( eg localhost ) for testing
# So identity-service.auth-server-url=http://<IP>:8999 should point to the right address

PR #248, opened 2019-05-07
RM-6680 Update RM version to the one compatible with ACS 6.0

cagache:
RM-6680 Update RM version to 3.0.1


aforascu:
Update pls the description and title and then you can do the merge

pom.xml
```diff
@@ -550,6 +550,12 @@
                 <version>${alfresco.glacier-connector.version}</version>
                 <type>amp</type>
             </dependency>
+            <dependency>
```
ancutam: @cagache ACS doesn't bring RM as a dependency by default. There is only a property added with the version, used for guidelines and for the integration testing.
So this dependency here needs to be removed.

PR #252, opened 2019-05-08
REPO-4369: Add org.apache.camel dependencies, version 2.23.2

ancutam:
Added `org.apache.camel dependencies`, version 2.23.2
After the upgrade, the test on the duplicate library failed, but it was not a valid case, so I did some changes to the script that does this duplication check:
* added a whitelist of library names, in order to fix the false positives that are reported for libraries that don't follow naming/versioning standards.
* refactored the script, for consistent naming style.

### DIFF
--- a/docker-alfresco/test/docker-alfresco-all-amps-test/scripts/checkLibraryDuplicates.sh
+++ b/docker-alfresco/test/docker-alfresco-all-amps-test/scripts/checkLibraryDuplicates.sh
@@ -2,24 +2,44 @@
 
 # Script used to check for duplicated libraries after applying all amps
 
-Libdir=$1
-lib_list=""
-echo "Scanning libraries from $Libdir"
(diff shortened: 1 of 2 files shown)


docker-alfresco/test/docker-alfresco-all-amps-test/scripts/checkLibraryDuplicates.sh
```diff
@@ -2,24 +2,47 @@
...
+               skip=true
+               break
+           fi
+       done
+       if [[ "$skip" = true ]]; then
+           echo "Skip $current_lib"
+           continue
+       fi
+       
+       for other_lib in $(gls --ignore="$current_lib" "$lib_dir"); do
```
ancutam: The proper command for linux is `ls` not `gls`, this was fixed on the 6.0.N branch, after the merge was done.

PR #253, opened 2019-05-09
REPO-4381 : added docker-compose for prometheus and grafana. 

anechifor:
REPO-4381 : added docker-compose for prometheus and grafana.  For test use only.

### DIFF
--- a/docs/micrometer/README.md
+++ b/docs/micrometer/README.md
@@ -4,7 +4,7 @@ and useful files.
 
 Alfresco repository now exposes an endpoint ```/alfresco/s/prometheus``` that is compatible with the 
 [Prometheus](https://prometheus.io/) scarping and point. 
-This end point is read-only and does not currently require authentication. 
+This end point is read-only and does currently require authentication using an alfresco user. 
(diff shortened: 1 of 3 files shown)


docs/micrometer/README.md
```diff
@@ -4,26 +4,26 @@ and useful files.
 
 Alfresco repository now exposes an endpoint ```/alfresco/s/prometheus``` that is compatible with the 
 [Prometheus](https://prometheus.io/) scarping and point. 
-This end point is read-only and does not currently require authentication. 
+This end point is read-only and does currently require authentication using an alfresco user. 
 Prometheus uses a pulling pattern on the specified end points and is able to provide 
 aggregation of time series events that help in monitoring our application.
```
andrei-rebegea: Add a few lines to say that users need to setup proper authentications in the authentication chain in alfresco and in Prometheus now and that Prometheus only supports a subset of the authentications supported by alfresco.
Maybe add a link here to :  https://prometheus.io/docs/prometheus/latest/configuration/configuration/

PR #258, opened 2019-05-10
Add a whitelist of library names, to the script that checks for duplicate libraries

ancutam:
Added a whitelist of library names, to the script that checks for duplicate libraries, empty for now, but could be useful to bypass false positives reported for libraries that don't follow naming/versioning standards (this is used on 6.0.N branch already).
Refactored the script, for consistent naming style.


aforascu:
LGTM

PR #294, opened 2019-06-26
Feature/local transform

alandavis:
Changes to support the local transformer.

Mainly to bring in the new 'misc' T-Engine and configure up the T-Router so that the repo can access it.


montgolfiere:
Thanks ! 

I wonder for future updates of T-Router/T-Engines (next EA &/or RC) whether there is some way to reduce the number of test.env / .env that need to be updated ?

PR #295, opened 2019-06-28
Using the wrong tag for alfresco-content-repository in docker-compose

alandavis:


docker-alfresco/test/docker-compose.yml
```diff
@@ -10,7 +10,7 @@ version: "3"
 
 services:
     alfresco:
-        image: quay.io/alfresco/alfresco-content-repository:feature-LocalTransform-6.2.0-SNAPSHOT
+        image: quay.io/alfresco/alfresco-content-repository:${ALFRESCO_TAG}
```
ancutam: `master` used to have the REGISTRY parameter as well in the repository path, instead of `quay.io/`. 
REGISTRY is empty (docker hub) by default and the build jobs override it to quay.io.

PR #312, opened 2019-07-09
REPO-4471 Update all-amps docker image for 6.0.N

andrei-zapodeanu-alfresco:


cristianturlica:
Changes look good to me, only thing to consider is the s3 properties should be removed and loaded from system variables.

andrei-zapodeanu-alfresco:
The removal of properties will be done in a different commit since it needs to be done in other branches at well.
This way, it will be easier to just cherry pick that commit across all branches.

docker-alfresco/Dockerfile
```diff
@@ -91,8 +91,8 @@ RUN mkdir -p ${TOMCAT_DIR}/conf/Catalina/localhost && \
     chmod g+rwx ${TOMCAT_DIR}/work && \
     chmod o-w ${TOMCAT_DIR}/work
 
-# To remote debug into this image add: EXPOSE 8000
-# Changes are also required to the docker-compose/docker-compose.yml file.
+# To remote debug into this image add: EXPOSE 8000	EXPOSE 10001
```
alandavis: ???

andrei-zapodeanu-alfresco: Removed

PR #345, opened 2019-07-23
MM-802: Re-enable Media Management amp

killerboot:
MM requires the following workarounds applied to be able to work with latest code:
* A full set of gytheio 0.9.1 libs as there is a number of them coming from alfresco-repository and also inside the amp. The libraries in the amp should be removed and the ones from war file should be used instead.
* camel-jackson should be updated separately as update of gytheio to 0.10 is not possible, see https://github.com/Alfresco/alfresco-repository/commit/36a6fcbc17a6985f4eb0c8ce81779d94823165f9
* Downgrade of metadata-extractor (TIKA dependency) to 2.11.0. See https://github.com/Alfresco/alfresco-repository/commit/b0f558c10219b1a1d4ad50b7c76298be841c80f2
It seems that MM is very fragile, updates to TIKA and it's dependencies should be done with that in mind.
* Including commons-lang 2.6 back into product. MM (and it's dependencies) should not use this old lib and move to commons-lang3, see https://issues.alfresco.com/jira/browse/REPO-4321
* Fixes to Hazelcast cache configs, see https://issues.alfresco.com/jira/browse/REPO-4503


montgolfiere:
As reviewed with @killerboot & @Epurashu this is the first stage to enable current MM 1.3.0 to "startup" on master (ACS 6.2.0-dev).

We will then continue to make further updates for MM 1.4.0 to target ACS 6.2.0 (including MM-802 and MM- 788 epic).

PR #418, opened 2019-09-10
WIP: MM-820: Update to Gytheio 0.10.x (=> 0.10.1)

montgolfiere:
- see corresponding release of alfresco-repository.jar (7.xx)
- https://github.com/Alfresco/alfresco-repository/pull/582


montgolfiere:
Note: branch build fails at mo. Not sure if related, but probably needs Repo version to be updated first. 

We also need to track any potential impact on MM & SS.

montgolfiere:
This is green with MM 1.4.0-TEST2 commented-out (in all-amps test) since that still relies in Gytheio class that has been removed (going from 0.9.1 to 0.10.1).

To avoid holding up Repo 6.2.0-A7, I will merge now. In the meantime, InT team will continue to fix MM (we have a 1.4.0-TEST3 in progress). We can then debug further using Repo 6.2.0-A7 (including Gytheio 0.10.1) once available.

PR #429, opened 2019-09-13
MM-820: Part 2 - update to MM 1.4.0-TEST4

montgolfiere:
- no longer relies on Gytheio's old JsonKeyClassDeserializer
- comment back into all-amps test :-)


montgolfiere:
FYI: I've updated to MM 1.4.0-TEST4 (which passed MM tests using Repo 6.2.0-A7 / Gytheio 0.10.1). 

Branch build is running ...  will merge if green (including all-amps startup test with MM).

PR #437, opened 2019-09-17
update sync service to 3.2.0-RC2

ciprianchelaru:


ancutam:
sync service to 3.2.0-RC2 was already updated on master

PR #449, opened 2019-09-20
Complete Migrating a Legacy Transformer Documentation.

Epurashu:
Add Legacy transformer transformOption and Pipeline config.

### DIFF
--- a/docs/migrating-a-legacy-transformer.md
+++ b/docs/migrating-a-legacy-transformer.md
@@ -2,10 +2,6 @@ For context see [Custom Transforms and Renditions](custom-transforms-and-renditi
 
 # Migrating a Legacy Transformer
 
-TODO
-* Talk about creating pipelines in json rather than Spring or via
-   values in transformer.properties.
-
 This section will describe how to migrate custom synchronous transformers
 created for Alfresco Content Repository (ACS) prior to version 6.2, to new
 asynchronous out of process T-Engines.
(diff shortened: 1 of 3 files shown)


docs/migrating-a-legacy-transformer.md
```diff
@@ -92,12 +89,39 @@ An example of this can be seen in the **HelloWorldController.java**.
...
+# content.transformer.alfresco-pdf-renderer.ImageMagick.priority=200
+# content.transformer.alfresco-pdf-renderer.ImageMagick.available=false
+# content.transformer.alfresco-pdf-renderer.ImageMagick.extension.ai.jpg.supported=false
 
-TODO
-* Talk about not needing to create a sub class of TransformOptions, or
-  the need to marshal and un marshal TransformOptions.
+```
+Pipeline definitions for Local Transformers are done via JSON rather than alfresco-global.properties
+see section [Configure a custom transform pipeline](https://github.com/Alfresco/acs-packaging/blob/master/docs/custom-transforms-and-renditions.md#configure-a-custom-transform-pipeline).
```
eknizat: We should add a priority to the example json pipeline..

eknizat: Also, should explain how the other options that you listed equate to the json format, for example the supported extension, in the json format you just list the ones you want to support. This means you don't have to explicitly say what you don't support.

PR #489, opened 2019-10-07
MM-828: Update ACS packaging version of MM from 1.4.0-RC3 to 1.4.0-RC5

montgolfiere:
- includes fix for MM-827 (fix MM Share AMP for Video Trim.& Image Crop)


montgolfiere:
Green branch build hence merging.

PR #541, opened 2019-10-28
REPO-4727 Add Legacy Transforms test to travis

PunchedChimera:
Added a docker-compose environment to enable rendition test.
Rendition tests added to travis.yml


PunchedChimera:
See travis build [#536](https://travis-ci.com/Alfresco/acs-packaging/builds/134230496)

PunchedChimera:
I followed the pattern used in bamboo. Using the variables: 
`LEGACY_TRANSFORM_SERVICE_ENABLED
LOCAL_TRANSFORM_SERVICE_ENABLED
TRANSFORM_SERVICE_ENABLED`
Setting the required service to true and those not required to false.

killerboot:
> I followed the pattern used in bamboo. Using the variables:
> `LEGACY_TRANSFORM_SERVICE_ENABLED LOCAL_TRANSFORM_SERVICE_ENABLED TRANSFORM_SERVICE_ENABLED`
> Setting the required service to true and those not required to false, allowing for only a single docker-compose file, instead of 3. If preferred 3 can be used.

I'm not sure that having one file is the best approach as:
* you need more scripting
* it is hard to understand how transformation is actually done
* the defaults in the .env file actually enables all of the transformation paths, I don't see what is the value of this config

@alandavis what is your opinion on this? are you planning to spend time in improving these tests?

PunchedChimera:
> * the defaults in the .env file actually enables all of the transformation paths, I don't see what is the value of this config

I'm unsure. This is how it is configured in [/docker-alfresco/test/.env](https://github.com/Alfresco/acs-packaging/blob/master/docker-alfresco/test/.env) used in [/docker-alfresco/test/docker-compose.yml](https://github.com/Alfresco/acs-packaging/blob/master/docker-alfresco/test/docker-compose.yml)

PunchedChimera:
@killerboot 
Separate docker-compose files are now used for each test job.

tests/environment/docker-compose-legacy-transforms-only.yml
```diff
@@ -24,28 +24,22 @@ services:
                 -Ddb.url=jdbc:postgresql://postgres:5432/alfresco
                 -Dalfresco.restApi.basicAuthScheme=true
                 -Dimap.server.enabled=true
+                -Dimap.server.port=1143
```
killerboot: do you need imap, ftp and etc for transform tests? If not it is better to remove those properties

killerboot:
Please check that the number of tests you run is the same as in Bamboo

tests/environment/docker-compose-legacy-transforms-only.yml
PunchedChimera: Removed imap and ftp.config and test pass

PR #575, opened 2019-11-14
WIP: REPO-4774: classpath collision commons-logging

amohammedalfresco:
- Exclude `commons-logging` classes that are being brought in by `spring-jcl`


killerboot:
Are you trying to exclude commons-logging library from the war?
If I build the war file with your changes I got both commons-logging and spring-jcl in it.
This is because dependencyManagement in the pom is not actually controlling if dependency is brought in or not. It is overriding the version if it is brought in.
See the relevant section here: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html

PR #607, opened 2019-11-27
REPO-4801 - Remove library org.springframework:spring-orm

antoniojfelix:
Regarding the issue REPO-4695, this is a library that can be removed according with the analysis done.


antoniojfelix:
Travis build failed.

PR #683, opened 2019-12-05
REPO-4841 - Remove library org.gytheio:gytheio-transform-worker-ffmpe…

antoniojfelix:
…g from acs-packaging project
Regarding the issue REPO-4695, this is a library that can be removed according with the analysis done.

### DIFF
--- a/war/pom.xml
+++ b/war/pom.xml
@@ -168,17 +168,6 @@
             </exclusions>
         </dependency>
         <dependency>
-            <groupId>org.gytheio</groupId>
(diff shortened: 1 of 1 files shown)


montgolfiere:
Merged without PR review ? Have we checked that this does not break existing Media Management 1.4.0. 

See also MM-802 and https://github.com/Alfresco/acs-packaging/pull/345/files (24/Jul/19).

antoniojfelix:
This PR was cancelled and the commit was reverted due to @montgolfiere input. Thanks for pointing that out. I will cancel the Jira ticket as well.

PR #848, opened 2020-01-15
MM-841 : Remove old commons-lang lib from Gytheio

DenisGabriela:
- upgrade gytheio
   - remove commons-lang

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,7 +52,7 @@
         <dependency.alfresco-jlan.version>7.1</dependency.alfresco-jlan.version>
         <dependency.alfresco-server-root.version>6.0.1</dependency.alfresco-server-root.version>
         <dependency.alfresco-api-explorer.version>6.3.0</dependency.alfresco-api-explorer.version>
-        <dependency.gytheio.version>0.11.2</dependency.gytheio.version>
+        <dependency.gytheio.version>0.11.3</dependency.gytheio.version>
         <dependency.alfresco-log-sanitizer.version>0.2</dependency.alfresco-log-sanitizer.version>
 
         <dependency.spring.version>5.2.3.RELEASE</dependency.spring.version>


DenisGabriela:
`All AMPs tests` Travis job keeps failing -> https://travis-ci.com/Alfresco/acs-packaging/builds/144515406. Another AMP(s) may still need the commons-lang 2.6 lib.

montgolfiere:
Anyway to find out via Travis which AMP (or AMPs). 

Otherwise, I guess we need to either run & repeat locally or scan the source code of all AMPs in the "all-amps" test list ?

https://github.com/Alfresco/acs-packaging/blob/master/docker-alfresco/test/docker-alfresco-all-amps-test/pom.xml

montgolfiere:
Anyway to find out via Travis which AMP (or AMPs). 

Otherwise, I guess we need to either run & repeat locally or scan the source code of all AMPs in the "all-amps" test list ?

https://github.com/Alfresco/acs-packaging/blob/master/docker-alfresco/test/docker-alfresco-all-amps-test/pom.xml

DenisGabriela:
Running the same commands as in the travis build I was able to find which module needs commons-lang:
`
mvn install -q -PenterpriseDocker
mvn install -q -f tests/tas-all-amps/pom.xml -DskipTests -Pall-tas-tests,prepare-wars-with-amps
docker-compose -f docker-compose-all-amps-test.yml up
`

Alfresco can't start because of the error below: 
 `Error creating bean with name 'FilePlanService_security' defined in class path resource [alfresco/module/org_alfresco_module_rm/rm-service-context.xml]: Invocation of init method failed; nested exception is java.lang.NoClassDefFoundError: org/apache/commons/lang/StringUtils`

Looks like RM uses the old commons-lang
I'll update my PR on acs-packaging to only update Gytheio's version and open a jira issue on RM.

montgolfiere:
Great, thanks.

PR #1050, opened 2020-03-17
Fix/repo 4906 veracode static scan

nithinnambiar:
Integrate Veracode static code scan with Travis.  Modified the travis file to add a new task to upload and scan alfresco.war file which generates a report on security flaws.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ branches:
   only:
     - master
     - /support\/.*/
-    - fix/REPO-4906_veracode-static-scan
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ sudo: required
 language: java
 jdk:
   - openjdk11
-env:
-  - DATE=`date +"%Y-%m-%d-%H%M"`
 
 services:
   - docker
@@ -21,8 +19,10 @@ branches:
   only:
     - master
     - /support\/.*/
+    - fix/REPO-4906_veracode-static-scan
 
 env:
+  - DATE=`date +"%Y-%m-%d-%H%M"`
   global:
     # Release version has to start with real version (6.3.0-....) for the docker image to build successfully.
     - RELEASE_VERSION=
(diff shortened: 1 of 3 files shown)


scripts/veracode/static_analysis.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
```
DenisGabriela: I see [this Travis job](https://travis-ci.com/github/Alfresco/acs-packaging/jobs/299049588) is green even though the scan could not be performed. Add `set -e` to exit the script when a command fails (returns a non-zero status).

scripts/veracode/static_analysis_init.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+echo "=========================== Starting Static Analysis Init Script ==========================="
+
```
DenisGabriela: Add `set -e` to exit the script when a command fails (returns a non-zero status).

.travis.yml
```diff
@@ -183,7 +182,14 @@ jobs:
         - ./tests/scripts/start-compose.sh ./tests/environment/docker-compose-all-amps-test.yml
         - ./tests/scripts/wait-for-alfresco-start.sh "http://localhost:8082/alfresco"
         - travis_wait 20 mvn install -f tests/tas-all-amps/pom.xml -Pall-tas-tests -Denvironment=default -DrunBugs=false
-
+    - name: "Veracode static code scan"
```
DenisGabriela: Similar to WhiteSource scan, we could run this only on master, support branches and on builds != pull_requests:
`     if: fork = false AND (branch = master OR branch =~ /support\/.*/) AND type != pull_request`.

nithinnambiar: The change has been made.

scripts/veracode/static_analysis_init.sh
nithinnambiar: The change has been made.

scripts/veracode/static_analysis.sh
nithinnambiar: The change has been made.

PR #1135, opened 2020-04-16
MNT-21417: LDAP Sync not working when making changes via admin UI

ancutam:
Add test for the LDAP sync job, after the subsystem is restarted.

I changed the existing travis job that used the LDAP env to run more suites, the rest api one and the new test I added for the ldap synchronization.


tests/tas-integration/src/test/java/org/alfresco/tas/integration/subsystems/ldap/LdapUserSynchronizationTest.java
```diff
@@ -0,0 +1,58 @@
...
+        {
+            String ldapEnabled = jmxBuilder.getJmxClient().readProperty("Alfresco:Type=Configuration,Category=Authentication,id1=managed,id2=ldap1", "ldap.authentication.active").toString();
+            Assert.assertEquals(ldapEnabled, Boolean.TRUE.toString(), String.format("Property ldap.authentication.active is [%s]", ldapEnabled));
+        });
+
+        String syncStatusPrevExecutionTime = jmxBuilder.getJmxClient().readProperty("Alfresco:Name=BatchJobs,Type=Synchronization,Category=manager", "SyncEndTime").toString();
+
+        STEP("2. Trigger the ldap users sync job");
+        jmxBuilder.getJmxClient().executeJMXMethod("Alfresco:Name=Schedule,Group=DEFAULT,Type=MonitoredCronTrigger,Trigger=syncTrigger", "executeNow");
+        Utility.waitToLoopTime(2, "Waiting for sync job to finish.");
```
aforascu, line 51: Is there any way we could use Utility.sleep instead (like on line 41) to check that the sync has happened instead of waiting 2 sec?

PR #1136, opened 2020-04-16
REPO-4859 : HTTP_UNAUTHORIZED instead of HTTP_FORBIDDEN for some CMIS

tzclucian:
- added TAS test


tzclucian:
@ancutam , @cristianturlica , @aforascu I've updated the test and it fails.
(I never thought I'd be so happy because a failed test).

It fails with exactly the same issue as in the JIRA ticket. Once we'll merge my fix, this should turn green.
https://travis-ci.com/github/Alfresco/acs-packaging/jobs/322092210

tzclucian:
The build is now failing because of the `UpdatePropertiesTests#deletedUserCannotUpdateFileProperties()` which expects a 401 status code / exception.

The issue is that when running with AIMS, after the [introduced fix](https://github.com/Alfresco/alfresco-repository/pull/932/files) it fails with a 403.

The test might not be in the best shape as when the call is made (just right after the user is deleted) the token appears to be valid.

Screenshot 2020-04-24 10 39 50

PR #1142, opened 2020-04-23
MNT-19887: Update Alfresco dependencies

CezarLeahu:

PR #1151, opened 2020-04-24
ACS-22 MNT-21514: Bump dependencies

Mykhail-Diachenko:


ancutam:
Update will be done in #1142

PR #1156, opened 2020-04-24
Fix adding logs in case alfresco is not starting

aforascu:
- fix by setting the proper container name

### DIFF
--- a/tests/scripts/wait-for-alfresco-start.sh
+++ b/tests/scripts/wait-for-alfresco-start.sh
@@ -27,7 +27,10 @@ if (("$COUNTER" < "$TIMEOUT")) ; then
 else
    echo "Waited $COUNTER seconds"
    echo "Alfresco Could not start in time."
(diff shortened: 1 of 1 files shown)


tests/scripts/wait-for-alfresco-start.sh
```diff
@@ -27,7 +27,10 @@ if (("$COUNTER" < "$TIMEOUT")) ; then
 else
    echo "Waited $COUNTER seconds"
    echo "Alfresco Could not start in time."
-   echo "Last 200 lines from alfresco.log on container $ALFCONTAINER:"
+   echo "All started containers:"	
+   docker ps -a	
+   ALFCONTAINER=`docker ps -a | grep environment_alfresco | awk '{ print $1 }'`
```
ancutam, line 32: Could you do here a `grep _alfresco` in case the name of the container is different in another context (like locally).

aforascu: Changed. Thanks

PR #1178, opened 2020-04-30
ACS-100 Update api-explorer to 6.2.1.1 into master

PunchedChimera:
Currently ACS 6.2.1-RC3 contains a reference to rest-api-explorer 6.3.0. This is wrong. There is a released 6.2.1 version but there have been more commits since. This PR updates master in order to use 6.2.1.1 version (released 29/04/20) which include these additional commits.


ancutam:
Why use 6.2.1.1 on master as well?

PunchedChimera:
> Why use 6.2.1.1 on master as well?

I think you're right, we'll leave it as 6.3.0.

PR #1218, opened 2020-05-15
APPS-136 : ACS Packaging all-amps startup tests fails (for MM 1.4.1-RC1)

DenisGabriela:
- configure s3 for mm

### DIFF
--- a/tests/environment/docker-compose-all-amps-test.yml
+++ b/tests/environment/docker-compose-all-amps-test.yml
@@ -47,6 +47,11 @@ services:
         -Ds3.bucketName=bamboo-s3-conn-default
         -Ds3.autoLowerCaseBucketName=true
 
+        -Dcontent.remote.default.contentRefHandler.source.s3.bucketRegion=us-east-1
+        -Dcontent.remote.default.contentRefHandler.target.s3.bucketRegion=us-east-1
+        -Dcontent.transformer.AwsElasticTranscoder.s3.bucketRegion=us-east-1
+        -Dcontent.transformer.Zencoder.s3.bucketRegion=us-east-1
+
         -Dconnector.az.account.name=${AZURE_STORAGE_ACCOUNT_NAME}
         -Dconnector.az.account.key=${AZURE_STORAGE_ACCOUNT_KEY}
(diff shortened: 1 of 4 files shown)


tests/environment/docker-compose-all-amps-test.yml
```diff
@@ -47,6 +47,11 @@ services:
         -Ds3.bucketName=bamboo-s3-conn-default
         -Ds3.autoLowerCaseBucketName=true
 
+        -Dcontent.remote.default.contentRefHandler.source.s3.bucketRegion=us-east-1
```
montgolfiere, line 50: I assume this also work with empty values ? If so, we should also raise a tech debt JIRA to add default (empty) properties to MM Repo AMP.

tests/tas-all-amps/pom.xml
```diff
@@ -195,14 +195,15 @@
                                             <overWrite>false</overWrite>
                                             <outputDirectory>${project.build.directory}/amps_share</outputDirectory>
                                         </artifactItem>
+                                        <!-- Incompatible with MM 1.4.1-RC1. Jira issue -> https://issues.alfresco.com/jira/browse/REPO-5133
```
montgolfiere, line 198: As we know, we should prioritise REPO-5133 asap to update S3/Glacier version (for ACS 6.2.1+) so that these does not remain commented-out for too long.

tests/environment/docker-compose-all-amps-test.yml
DenisGabriela: It is not working with empty values -> https://travis-ci.com/github/Alfresco/acs-packaging/jobs/334631259, so I assume we need these default configs.

tests/tas-all-amps/pom.xml
DenisGabriela: Yes, I agree.

CezarLeahu:
Assuming we merge the ACS-225 workaround (https://github.com/Alfresco/media-management/pull/450), we should just release another MM RC (**MM-1.4.1-RC2**) and upgrade to that without any additional changes/fixes.

DenisGabriela:
Makes sense. I'll close this PR without merging.

PR #1364, opened 2020-06-25
ACS-370: Newer metadata-extractor (2.13.0)

ancutam:
Add direct dependency in acs-packaging


ancutam:
Already done in #1368

PR #1373, opened 2020-06-25
APPS-222 : Create Share 6.2.2-RC3

alexbalmus:
- acs-packaging PR for Share 6.2.2-RC3


montgolfiere:
Note: I've closed this PR without merge since we now have final Share 6.2.2 applied to support/SP/6.2.N

PR #1374, opened 2020-06-26
Fix/mnt 21602 script node return invalid donwload url

Epurashu:
Revived DownloadContentServlet with respective base and helper classes.

Removed inaccessible code and methods from the original code (as tightly coupled to Alfresco Explorer).

Created `public static void prettyPrintError(HttpServletRequest req, HttpServletResponse res, String message, int status, String actualCause)` as there past error pages are not available anymore, this is done to avoid printing of stacktrace into the browser.

Removed leftover `rg.alfresco.web.app.servlet.UploadContentServlet`, `org.alfresco.web.app.servlet.DownloadRawContentServlet` and `org.alfresco.web.app.servlet.GuestDownloadContentServlet` from web.xml as those servlet implementation does not exist.

### DIFF
--- a/war/src/main/java/org/alfresco/web/app/Application.java
+++ b/war/src/main/java/org/alfresco/web/app/Application.java
@@ -1,0 +1,238 @@
+/*
+ * #%L
+ * Alfresco Repository WAR Community
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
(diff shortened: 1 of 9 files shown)


killerboot:
According to this [comment](https://issues.alfresco.com/jira/browse/MNT-21602?focusedCommentId=795575&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-795575)
the chose approach is wrong.

Epurashu:
@killerboot thank you for your reply.

There are 2 point in the `Expected Behaviour` of that comment:

1. Download URL strings/values returned using ScriptNode's "url" & "downloadUrl" methods must work similarly to how it was working until 5.2.X
2. This download mechanism is no longer available from 5.0 onwards. There is no functional change between 5.2.x up to the latest version.

Point 1 is working as expecting, the ScriptNode returns the same url at least since 4.2.X (I tested on 4.2.8 but I think it is avaialble for older versions as is tightly coupled with the old _Alfresco Explorer_).

Regarding point 2, there were some discussions here, the download mechanism (the one that this commit brings back)  was partially working from 5.0 onwards. The login page and error pages were not available (I made a pretty print function to not display stacktrace anymore) but there are 2 use cases where this endpoint is working:

1. Providing a ticket into the url: `/d/d/{storeType}/{storeId}/{nodeRef}/{nodeName}?ticket={validTicket}
2. Using External Authentication.

killerboot:
The ticket can potentially work using "alf_ticket": https://docs.alfresco.com/6.2/tasks/ws-specify-user-identity.html using standard authentication filters. Potentially it is possible to rewrite the URLs to V1 APIs during runtime.
In my opinion this bunch of copied code is mostly "dead code", this increases maintenance load. If this is really brought back it needs to have a very high test coverage.

Epurashu:
I agree with you, the code has some bugs, I will create a ticket to remove decide on EOL/remove this code from 7.X, refactor scriptNode API and docs changes accordingly (this is @skopf suggestion).

Regarding the rewrite, I've asked Support if the customer wants to use V1 RestAPI instead of this Servlet or if the customer agrees with ScriptNodeAPI returning a V0/V1 download url (working the same). They are not happy with those proposals.
The customer expects `DownloadContentServlet` to work on 6.X as it was working on 5.2.X.

Epurashu:
Closed, different approach merged in this [PR#1392](https://github.com/Alfresco/acs-packaging/pull/1392)

PR #1392, opened 2020-06-29
Fix/mnt 21602 script node return invalid download url redirect

Epurashu:
Revive DownloadContentServlet.java
Remove doHEAD method from `DownloadContentServlet`
doGET method now redirects to V1 REST - `GET /nodes/{nodeId}`
Removed leftover `org.alfresco.web.app.servlet.UploadContentServlet`, `org.alfresco.web.app.servlet.DownloadRawContentServlet` and `org.alfresco.web.app.servlet.GuestDownloadContentServlet` from web.xml as those servlet implementation does not exist.


Epurashu:
@killerboot thank you, I will add them now.

war/src/main/java/org/alfresco/web/app/servlet/DownloadContentServlet.java
```diff
@@ -0,0 +1,124 @@
...
+      StoreRef storeRef = new StoreRef(URLDecoder.decode(t.nextToken()), URLDecoder.decode(t.nextToken()));
+      String id = URLDecoder.decode(t.nextToken());
+
+      // build noderef from the appropriate URL elements
+      NodeRef nodeRef = new NodeRef(storeRef, id);
+
+      // build redirect URL to V1 GET /nodes/{nodeId}
+      String redirectUrl = String
+          .format("%s://%s:%s%s/api/-default-/public/alfresco/versions/1/nodes/%s/content?attachment=%b",
+              req.getScheme(), req.getServerName(), req.getServerPort(), req.getContextPath(), nodeRef.getId(), isAttachment);
```
skopf, line 116: This doesn't work. The req.getScheme(), ... values are not what we need in the redirect URI

Epurashu: Thank you for spotting this, I have removed the `req.getScheme()`, `req.getServerName()` and `req.getServerPort()` from the `redirectUrl`

Epurashu:
@killerboot I have added `verifyDownloadContentServletRedirect()` inside the `TAS-Integration` project. The test turned out a little bigger, I`m thinking of moving this tests to it's own class and splitting it into smaller ones, what do you think ?

killerboot:
I think the tests would be better in a separate class in a TAS REST API project instead of integrations.
Overall the tests look ok to me.

war/src/main/java/org/alfresco/web/app/servlet/DownloadContentServlet.java
```diff
@@ -0,0 +1,124 @@
...
+      {
+         throw new IllegalArgumentException("Attachment mode is not properly specified: " + requestURI);
+      }
+
+      // assume 'workspace' or other NodeRef based protocol for remaining URL elements
+      StoreRef storeRef = new StoreRef(URLDecoder.decode(t.nextToken()), URLDecoder.decode(t.nextToken()));
+      String id = URLDecoder.decode(t.nextToken());
+
+      // build noderef from the appropriate URL elements
+      NodeRef nodeRef = new NodeRef(storeRef, id);
```
skopf, line 111: Won't work like this. You do not need to build a NodeRef if you are only using the ID part.

And we need to check beforehand that this request is for workspace://SpaceStore/....
other requests, like for archived documents, cannot be fulfilled by this V1 rest api

PR #1393, opened 2020-06-29
ACS-250: Test S3 and Glacier Connectors' compatibility with 6.2.2

ancutam:
Applied S3 and Glacier Connectors and commented out MM.
This is a test PR, to make sure the build is green with this combination of amps.


ancutam:
Green build https://travis-ci.com/github/Alfresco/acs-packaging/builds/173457684

PR #1427, opened 2020-07-07
REPO-5235: Unfork tika

eknizat:


pom.xml
```diff
@@ -177,6 +177,11 @@
 
     <dependencyManagement>
         <dependencies>
+            <dependency>
```
montgolfiere, line 180: Why did we need to add this ?

eknizat: It caused a test failure when checking for duplicates.

https://issues.alfresco.com/jira/browse/REPO-5235?focusedCommentId=808373&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-808373

PR #1464, opened 2020-07-15
ACS-222: tests jobs using AIMS fail

tiagosalvado10:
From the analysis done in ACS-222, adding an extra interval (before tests start) seems to prevent the jobs to fail - not only the AIMS jobs but also the other ones that already failed before ACS-222 (even that not very often).

### DIFF
--- a/tests/scripts/wait-for-alfresco-start.sh
+++ b/tests/scripts/wait-for-alfresco-start.sh
@@ -24,6 +24,7 @@ if (("$COUNTER" < "$TIMEOUT")) ; then
    t1=$(date +%s)
    delta=$((($t1 - $t0)/60))
    echo "Alfresco Started in $delta minutes"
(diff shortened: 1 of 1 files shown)


tests/scripts/wait-for-alfresco-start.sh
```diff
@@ -24,6 +24,7 @@ if (("$COUNTER" < "$TIMEOUT")) ; then
    t1=$(date +%s)
    delta=$((($t1 - $t0)/60))
    echo "Alfresco Started in $delta minutes"
+   sleep 180
```
ancutam, line 27: Please add a comment here or modify the message to say that Alfresco started successfully in x minutes, waiting for all the containers to initialise", or something along that. And a message after the 3 minutes end.

Also add a short description to the PR.

PR #1551, opened 2020-08-05
MNT-21614: LDAP sync communication error leads to group members deletion

amohammedalfresco:
- Bump alfresco-repository version and add a test


amohammedalfresco:
> Does the test fail without the fix?

It doesn't fail without the fix.  Actually, it's quite hard to reproduce the exact reported issue.  This test only catches one of the exception.  It's actually not even related to the bug, but thought to add this expected exception.

The problem is the docker-compose environment cannot be manipulated for reproducing other than either a stop of ldap container or using a non-existent port (I chose to just use a non-existent port for the duration of test).

amohammedalfresco:
> Please make sure the test revert the invalid config after it runs. If it doesn't, then the consecutive tests may fail unexpectedly.

Sure.  I've just guarded the test to make sure port reversion is as expected.

PR #1670, opened 2020-09-21
APPS-510-bump to aos 1.3.2-A3

pieCit87:


pieCit87:
Im going to close this PR in order to set bumps in different orders

PR #1678, opened 2020-09-22
Bump to MM 1.5.0-M2

pieCit87:
Upgrade of dependency, common-services, commons-lang


pieCit87:
Closing because upgraded ACS packaging from dependabot alert PR https://github.com/Alfresco/acs-packaging/pull/1687

PR #1686, opened 2020-09-23
MNT-21859 Embed old keystore in docker images

killerboot:
Contents of the PR:
* Copy the old keystore into the docker image from distribution zip
* Configure docker environments to use old keystore's config as JAVA_OPTS
* Remove redundant enterpriseDocker maven profile as it is the same as fullBuild because of the dependency on the distribution zip.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ jobs:
     - name: "REST API TAS tests part1"
       jdk: openjdk11
       install:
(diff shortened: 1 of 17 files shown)


killerboot:
Added @alandavis to make sure this doesn't break the new repo builds.

gavincornwell (approved):
LGTM

skopf (approved):
lgtm

PR #1692, opened 2020-09-24
Add in all amps : ai outlook sap

eromano:
pipeline-all-amps/ needs to have in order to test all the AMPS that PASS is going to use the following AMPS:

ai-repo
emailclient outlook
SAP


alandavis (approved):
LGTM, but could you please check if there is a space/tab issue in the pom.xml files.

PR #1787, opened 2020-11-09
ACS-595 Direct Access URLs design document

CezarLeahu:
Please use this link to view the design document: [https://github.com/Alfresco/acs-packaging/blob/ACS-595_direct-access-urls-design/docs/design-doc-direct-access-links.md](https://github.com/Alfresco/acs-packaging/blob/ACS-595_direct-access-urls-design/docs/design-doc-direct-access-links.md).


stevenblair-al:
@CezarLeahu I've raised a question in slack for alignment with ACS-540 from Irving I would like answered please.

skopf:
Design looks really good. Some small comments:

https://github.com/Alfresco/acs-packaging/blob/ACS-595_direct-access-urls-design/docs/design-doc-direct-access-links.md#system-wide-configuration-acs
"The ACS system-wide configuration take precedence over the other more specific settings"
Actually, the system wide setting does not always take precedence. For example, if I have set a lower maxTimeout for a specific content store, then DAUs of this CS cannot exceed this limit even if the system wide setting is hight.
Also the connector specific connector.s3.directAccessUrl.defaultExpiryTimeInSec will override the system wide default.
You could change this sentence to something like "The individual configuration parameters are evaluated in the order defined below" or similar.

Also here
https://github.com/Alfresco/acs-packaging/blob/ACS-595_direct-access-urls-design/docs/design-doc-direct-access-links.md#configuration-priorities
"In terms of expiry times (both default & max), the rule of thumb is: the system-wide settings take precedence over the content-store-specific or rest-api-specific configuration."
The is in conflict with the details described above. For example, if I have a system wide max of 5 min and a s3 specific max of 1 min, then DAUs for S3 cannot exceed 1 min timeout. Same for the defaults.
Otherwise, the content store specific would not make any sense. The system wide setting is mandatory, and if it always overwrites the content store specific setting, then this content store specific setting would never be used.

gavincornwell:
Firstly, you've done an awesome job of producing a very detailed design document makes it so much easier to understand the problem domain.

It looks really good to me. 

However, it would be good to see the OpenAPI spec in there too and also have a question about discoverability. Is there an efficient way to determine whether a particular node is capable of producing a DAU? Rather than expecting clients to try and handle an error it would be nice if we could return back a flag (maybe via the include mechanism) as part of the node response so that clients could just hide the link/button for the individual files that don't support DAUs.

docs/design-doc-direct-access-links.md
```diff
@@ -0,0 +1,813 @@
...
+    EP -> SV ++ : Request DAU
+    return DAU
+    return DAU
+end
+
+@enduml
+```
+
+### UI Client Integration example
+For downloading a file from ACS, a client application could implement a logic that first attempts
```
montgolfiere, line 729: I think this needs a detailed design section for ADW itself, ie. exposing this feature end-to-end, including all aspects of the UX, ie. including configuration &/or discovery (from ADW) and whether it will be for Download only (including versions) or also potentially Previews (ie. renditions). 

This should enable a feature team to potentially deliver the complete end-to-end feature, eg.. S3 Connector, Repository APIs and the ADW changes. Ideally also the Azure Connector as well.

ancutam:
This was moved to another repo.

PR #1791, opened 2020-11-18
ACS-857 : Azure Connector 1.2 Alpha Release [master]

tzclucian:
- updating azure


CezarLeahu (approved):

aforascu (changes requested):
I think we need to also uncomment the test - https://github.com/Alfresco/acs-packaging/blob/691cb3e42d3c9abc43d1e77136920cb0882a8d19/tests/tas-all-amps/src/test/java/org/alfresco/rest/discovery/DiscoveryTests.java#L74

aforascu (approved):

PR #1804, opened 2020-12-04
Ability to specify a suffix and publish 2 images with and without ags

mauriziovitale:


mauriziovitale:
going to close the issue as won't fix

PR #1805, opened 2020-12-04
ACS-999 Run ACS on docker with a configurable set of AMPs

mteodori:
See: ACS-999

Added `ALFRESCO_AMPS` env variable to select which AMPs to load when running the image for both repo and share.

*NB* it requires ROOT access due to some unexpected action from the install tool outside the webapp dir that need to be investigated and removed, but this is not an issue at the moment as this docker image is meant to support CI/CD pipelines, should we decide to extend and have this image as the single one considering it covers any combination use case, it would be worth to create an issue to fix that


eromano (approved):

eromano:
Nice idea!

tests/pipeline-all-amps/repo/Dockerfile
```diff
@@ -44,5 +40,14 @@ RUN chgrp -R Alfresco ${TOMCAT_DIR} && \
   chmod -R g+r ${TOMCAT_DIR}/webapps
 
 #Use the alfresco user
-USER alfresco
+#USER alfresco
+
+ENV TOMCAT_DIR=${TOMCAT_DIR}
+
+ENV ALFRESCO_AMPS="alfresco-aos-module alfresco-device-sync-repo alfresco-glacier-connector-repo alfresco-googledrive-repo-enterprise alfresco-s3-connector alfresco-saml-repo alfresco-share-services"
```
nithinnambiar, line 47: There is also an alfresco-azure-connector amps which is alternative to s3-connector. Is it possible to to add it?

tests/pipeline-all-amps/docker-entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+echo Previously installed AMPs in $ALFRESCO_WEBAPP_DIR
+INSTALLED_AMPS=$(java -jar $ALFRESCO_MMT_JAR list $ALFRESCO_WEBAPP_DIR)
+echo "$INSTALLED_AMPS"
+INSTALLED_AMPS=$(echo "$INSTALLED_AMPS" | grep ^Module | cut -f2 -d' ' | tr -d "'")
+
+echo Uninstall previously installed AMPs in $ALFRESCO_AMPS_DIR
+for AMP in $INSTALLED_AMPS
+do
+	echo Uninstalling $AMP
+	java -jar $ALFRESCO_MMT_JAR uninstall $AMP $ALFRESCO_WEBAPP_DIR
```
gavincornwell, line 20: Historically this hasn't been very reliable, unfortunately some AMPs overwrite files and uninstalling can cause issues unless you have a backup, which won't be the case here.

Some of the AMPs pre-installed are also required i.e. share-services and trash can cleaner. So do you really need to do this?

gavincornwell (commented):
I think this is OK for a test pipeline, other than my inline comment about uninstalling AMPs.

tests/pipeline-all-amps/repo/Dockerfile
```diff
@@ -44,5 +40,14 @@ RUN chgrp -R Alfresco ${TOMCAT_DIR} && \
   chmod -R g+r ${TOMCAT_DIR}/webapps
 
 #Use the alfresco user
-USER alfresco
+#USER alfresco
```
tpage-alfresco, line 43: This can be deleted

tests/pipeline-all-amps/docker-entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+	cp $ALFRESCO_AMPS_DIR/${AMP}* $TMP_ALFRESCO_AMPS_DIR
+done
+java -jar $ALFRESCO_MMT_JAR install \
+	$TMP_ALFRESCO_AMPS_DIR $ALFRESCO_WEBAPP_DIR -directory -nobackup -force -verbose
+unset TMP_ALFRESCO_AMPS_DIR
+
+echo Verify requested AMPs $ALFRESCO_AMPS have been installed
+java -jar $ALFRESCO_MMT_JAR list $ALFRESCO_WEBAPP_DIR
+
+exec "$@"
```
tpage-alfresco, line 41: Is it worth switching back to the alfresco user at this point?

tests/pipeline-all-amps/repo/run-repo.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -e
+
+PROJECT_NAME=repo
+[[ "$BUILD_ENABLED" == "true" ]] && mvn clean install
+env \
+	DOCKER_IMAGE_REPO=alfresco/alfresco-pipeline-all-amps-$PROJECT_NAME \
+	../run.sh "$@"
```
tpage-alfresco, line 9: It took me a while to work out what this script was doing, and I'm still not sure who we're expecting to use it.  Please could you add a comment (or `-h` option would be better, but it slightly interferes with the use of `$@`) to explain how to use it?

tests/pipeline-all-amps/run.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e
+
+IMAGE_TAG=7.0.0-M2
```
tpage-alfresco, line 5: Can we get this from a pom file somewhere?  Is there a reason why we're using 7.0.0-M2 rather than 7.0.0-A14?

tests/pipeline-all-amps/docker-entrypoint.sh
alandavis: I too think it would be far better to start with the simplest Repo image and always ADD to it. Uninstall is really not supported and we will always be asking is it really the same as just adding onto the base.

tests/pipeline-all-amps/run.sh
```diff
@@ -0,0 +1,15 @@
...
+
+sed -e "s@\${image.tag}@$IMAGE_TAG@" Dockerfile > target/Dockerfile
+cp ../docker-entrypoint.sh target/
+
+docker rmi -f $DOCKER_IMAGE_REPO
+docker build -f target/Dockerfile -t $DOCKER_IMAGE_REPO .
+echo http://localhost:8080$BASE_PATH
+# --user 1000:1000 \
```
tpage-alfresco, line 13: I guess we can delete this.

tests/pipeline-all-amps/run.sh
```diff
@@ -0,0 +1,15 @@
...
+docker run --rm -it \
+  --publish 8080:8080 $DOCKER_IMAGE_REPO "$@"
```
tpage-alfresco, line 15: I'm not sure what the expected use case for this script is, but it seems to lose some value by picking 8080 for both repo and share.

tests/pipeline-all-amps/docker-entrypoint.sh
alandavis: Another approach, I have been thinking about is putting the various amps into a different directory and the providing a way to add selected ones to the war. At least then the image contains everything to start with.

tpage-alfresco (commented):
I like the idea of dynamically picking amps at runtime and the docker side of things. I didn't understand why the "run" scripts were added and I think they could benefit from some comments and/or usage instructions.

killerboot:
The concerns about this approach are:
* The startup time of the repository container will increase, it is already too slow. This will affect the readiness probes.
* The installation of AMPs can break the startup. Meaning that the container will just die. In k8s it will endlessly restart without any evidence why. This means that the end result of installation (the container with amps installed) needs to be tested prior to be used in one pipeline or anywhere else, but it is not possible as the list of AMPs is a variable here.
So I would recommend to think about a different way of doing that.

tests/pipeline-all-amps/docker-entrypoint.sh
mteodori: I wasn't aware of the base repo image, will use that one and remove the uninstall workaround

tests/pipeline-all-amps/repo/Dockerfile
mteodori: ideally I would like to put it back, but I haven't understood why the alfresco-mmt needs write access outside of the target webapp

tests/pipeline-all-amps/docker-entrypoint.sh
mteodori: not sure I can do sudo/su, it would work only if running as root user

tests/pipeline-all-amps/repo/run-repo.sh
mteodori: it was helpful for me to test as it uses plain docker rather than magic from maven plugins

tests/pipeline-all-amps/run.sh
mteodori: I was just testing the replacement in local dev, not used by pipeline but ideally this should be used rather than the maven plugin that doesn't work with latest docker desktop on my mac

tests/pipeline-all-amps/run.sh
mteodori: meant for local testing, ideally this should work with the unprivileged user commented but this requires the alfresco-mmt to be fixed in order not to touch anything outside the webapp

mteodori: I should create an issue for that as the code is not straightforward for something that in the end should just unpack a zip in a dir

tests/pipeline-all-amps/run.sh
mteodori: local testing with standard tool but we can do such change if we want to use them as part of some env test

tests/pipeline-all-amps/repo/Dockerfile
mteodori: > The concerns about this approach are:
> 
> * The startup time of the repository container will increase, it is already too slow. This will affect the readiness probes.
> * The installation of AMPs can break the startup. Meaning that the container will just die. In k8s it will endlessly restart without any evidence why. This means that the end result of installation (the container with amps installed) needs to be tested prior to be used in one pipeline or anywhere else, but it is not possible as the list of AMPs is a variable here.
>   So I would recommend to think about a different way of doing that.

when testing with the run scripts, unzip execution time was negligible so I am not expecting it affects the probes
probes should already be configured with a max restart anyway and the log would show any issues, the logic executed is very limited and all dependencies are inside so probably this would be a problem only when we pick the other issue I had in mind that is to install from a volume any custom non alfresco AMPs a customer might have

happy to review alternatives if any but for me having a docker image with all the supported Alfresco AMPS and the opportunity to enable the ones you need is already miles better than what we have now where anyone seem to create their own testing docker image with a custom combination of amps

tpage-alfresco:
@mteodori Thanks for the replies about the "run" scripts. It sounds like these are not an integral part of this change, and additionally aren't in a suitable state to be widely useful or maintained.  I suggest that we just delete them from this PR if you don't intend to add some documentation to explain how to use/maintain them.

Other than these files I'm happy with the overall approach for test environments.

tests/pipeline-all-amps/repo/Dockerfile
mteodori: > There is also an alfresco-azure-connector amps which is alternative to s3-connector. Is it possible to to add it?
what are the artifacts coordinates?

mteodori: neve mind, I've found them all https://github.com/Alfresco/acs-packaging/blob/master/tests/tas-all-amps/pom.xml#L115-L313

tests/pipeline-all-amps/docker-entrypoint.sh
mteodori: the base image seem to be built only for latest, so it looks like it can't be used

mteodori: the amp jars are all included in a dir but installed in the webapp only if named in the ALFRESCO_AMPS whose default value is the full list so I guess @alandavis that's already what you would like

mteodori:
> @mteodori Thanks for the replies about the "run" scripts. It sounds like these are not an integral part of this change, and additionally aren't in a suitable state to be widely useful or maintained. I suggest that we just delete them from this PR if you don't intend to add some documentation to explain how to use/maintain them.
> 
> Other than these files I'm happy with the overall approach for test environments.

@tpage-alfresco makes sense, will probably add a README as they are useful for local development or just delete

tpage-alfresco (approved):
Thanks for the docs and updating the port numbers Marcello. LGTM

PR #1821, opened 2020-12-16
OPSEXP-675: Add API Explorer back into pipeline-all-amps

gavincornwell:
Since https://github.com/Alfresco/acs-packaging/pull/1805/files was merged for ACS-999 the /api-explorer has disappeared from the pipeline-all-amps image.

Presumably this has been caused by the Dockerfile change where we now extend from "alfresco-enterprise-repo-base" instead of "alfresco-content-repository", this PR adds the repo API Explorer back using the same method as the RM API Explorer.


eromano (approved):

eromano:
LGTM

PR #1822, opened 2020-12-17
SEARCH-2577 remove direct dependency from ES client, update ES image …

robfrank:
remove direct dependency from ES client in favour of the transitive one provide by elastic search connector.

In this way, updating the version on connector side would be enough

Only the .env file for the docker compose should be updated manually


jesty (approved):
LGTM

PR #1829, opened 2021-01-06
ACS-1077 push pipeline images in install phase

mteodori:
As publishing images happens on the maven install phase for all the images but the pipeline ones, moving push from deploy to install would fix ACS-1077


alandavis (approved):
Let's give it a go.

PR #1838, opened 2021-01-13
Fix Search TAS tests

DenisGabriela:
- change elastic search connector tag as the tests fail with the latest docker images


tpage-alfresco (approved):
Thanks!

alessandrobenedetti (approved):
according to our analysis, with @tpage-alfresco and @DenisGabriela , we should not have a floating tag for the ES connector

cristianturlica (approved):

ancutam (approved):

PR #1846, opened 2021-01-19
Search-2658 fix Elasticsearch tests

jesty:


tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchTests.java
```diff
@@ -176,14 +182,14 @@ public void searchCanFindAFileAsOwner() throws Exception
...
+            TestGroup.SEARCH }, executionType = ExecutionType.REGRESSION, description = "Verify that Elasticsearch search works as expected when the user can search a file because he is the owner.")
     @Test(groups = { TestGroup.SEARCH }, enabled = false)
-    public void searchCanFindAFileOnMultipleSites() throws Exception
+    public void searchCanFindAFileOnMultipleSitesWithOwner() throws Exception
     {
         Utility.sleep(1000, 10000, () -> {
             SearchRequest query = new SearchRequest();
             RestRequestQueryModel queryReq = new RestRequestQueryModel();
-            queryReq.setQuery("This");
+            queryReq.setQuery("TEST");
```
jesty, line 192: I changed the query because "this" could be a stop word.

killerboot (changes requested):
Please remove the necessity to cleanup the index. Use random data.
Also I don't think that the goals of SEARCH-2658 are met. There is still code in this test that creates mapping. the test should not depend on alfresco-elasticsearch-shared lib.
If this work depends on SEARCH-2632, then we should pause this work and do it properly once SEARCH-2632 is finished.

jesty:
> SEARCH-2632

I add a flag on SEARCH-2658 in order to highlight that we are waiting for SEARCH-2632

jesty:
> Please remove the necessity to cleanup the index. Use random data.
Because we are testing search queries I guess it's a bit difficult to use random data. 
Because the test on Travis will start the docker container before running the test I guess we can remove the cleanup, but this can be annoying locally.

> Also I don't think that the goals of [SEARCH-2658](https://alfresco.atlassian.net/browse/SEARCH-2658) are met. There is still code in this test that creates mapping. 

I removed any mapping creation.

> the test should not depend on alfresco-elasticsearch-shared lib.
I removed any dependency from the shared lib, in the past, it was useful because we are creating the mapping manually, but now we don't need it anymore.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchTests.java
```diff
@@ -80,20 +82,20 @@
...
      *  - Documents: FILE_2_NAME (owner: userSite2)
      *  
      */
     @BeforeClass(alwaysRun = true)
-    public void dataPreparation()
+    public void dataPreparation() throws IOException
     {
         serverHealth.assertServerIsOnline();
 
         elasticClient = new RestHighLevelClient(RestClient.builder(new HttpHost("localhost", 9200, "http")));
```
killerboot, line 97: The connection parameters should not be hardcoded.

killerboot:
> Because we are testing search queries I guess it's a bit difficult to use random data.
Because the test on Travis will start the docker container before running the test I guess we can remove the cleanup, but this can be annoying locally.
Obviously, if you have some smart solution or example in order to use random, I will be happy to do it.

I removed the index clean up and the tests run just fine multiple time. I think it works, because you are using random users and private sites.

tests/tas-elasticsearch/pom.xml
```diff
@@ -51,9 +51,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.alfresco</groupId>
-            <artifactId>alfresco-elasticsearch-shared</artifactId>
-            <version>${alfresco-elasticsearch-shared.version}</version>
+            <groupId>org.elasticsearch.client</groupId>
```
killerboot, line 54: Please remove this dependency and the tests that rely on it.

jesty: done

jesty:
> > Because we are testing search queries I guess it's a bit difficult to use random data.
> > Because the test on Travis will start the docker container before running the test I guess we can remove the cleanup, but this can be annoying locally.
> > Obviously, if you have some smart solution or example in order to use random, I will be happy to do it.
> 
> I removed the index clean up and the tests run just fine multiple time. I think it works, because you are using random users and private sites.

You are right

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchTests.java
jesty: This line doesn't exist anymore, it is removed after the Elasticsearch client removal.

killerboot (approved):

PR #1849, opened 2021-01-20
The Single pipeline jobs should not have any commented out AMPs.

alandavis:
Also removed the DTE AMPs as there should not be a Repo on in 7.0.0 and we don't know about Share.

As a result of comments of this PR, desktop sync and AGS (rm) have been reintroduced to the DiscoveryTests and removed the commented out DTE. More tests will fail, but the code is more complete once these AMPs are fixed.

### DIFF
--- a/tests/pipeline-all-amps/repo/pom.xml
+++ b/tests/pipeline-all-amps/repo/pom.xml
@@ -25,6 +25,7 @@
                         </goals>
                         <configuration>
                             <artifactItems>
+                                <!-- Do not comment out, as this will cause problems in the Single Pipeline. -->
                                 <artifactItem>
                                     <groupId>org.alfresco.saml</groupId>
                                     <artifactId>alfresco-saml-repo</artifactId>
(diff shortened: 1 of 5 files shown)


killerboot (commented):
Can you please also fix org.alfresco.rest.discovery.DiscoveryTests#getRepositoryInstalledModules ?
It should fail if the list of amps installed is not exactly as it is specified in the tests/tas-all-amps/pom.xml. Potentially we can use a similar test against the pipeline image to act as a double check of the modification of amp list.

killerboot (approved):

cristianturlica (approved):

tests/pipeline-all-amps/repo/pom.xml
```diff
@@ -112,32 +123,25 @@
...
-<!--                                    <type>amp</type>-->
-<!--                                    <overWrite>false</overWrite>-->
-<!--                                    <outputDirectory>${project.build.directory}/amps</outputDirectory>-->
-<!--                                </artifactItem>-->
+                                </artifactItem>
+                                <!-- Do not comment out, as this will cause problems in the Single Pipeline. -->
+                                <!-- TODO fix test failure APPS-702 -->
+                                <artifactItem>
+                                    <groupId>org.alfresco</groupId>
+                                    <artifactId>alfresco-governance-services-enterprise-repo</artifactId>
```
ancutam, line 139: While cleaning this up, `alfresco-governance-services-enterprise-repo` amp appears two times :)

alandavis: Thanks Ancuta, did not see the comment for some reason. Has been included in the next commit.

PR #1857, opened 2021-01-27
Use docker tag to follow the RFC on docker tags

eromano:
https://alfresco.atlassian.net/wiki/spaces/CTO/pages/26543757/RFC+Release+tags+for+Docker+images

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -195,7 +195,7 @@ jobs:
     - name: "Update latest and Single Pipeline <acs>-<build> images"
       stage: docker_latest
       # Remove the pipeline profile to avoid sending an image tagged with the build number to the single pipeline
-      script: travis_retry travis_wait 30 mvn -B -V clean install -DskipTests -Dmaven.javadoc.skip=true -Dbuild-number=${TRAVIS_BUILD_NUMBER} -Pinternal,pipeline
+      script: travis_retry travis_wait 30 mvn -B -V clean install -DskipTests -Dmaven.javadoc.skip=true -Dbuild-number=${TRAVIS_BUILD_NUMBER} -Dbranch=${TRAVIS_BRANCH} -Pinternal,pipeline
 
     - name: "Release and Copy to S3 Staging Bucket"
       stage: release
--- a/tests/pipeline-all-amps/pom.xml
+++ b/tests/pipeline-all-amps/pom.xml
@@ -61,7 +61,7 @@
             <id>internal</id>
             <!-- publishes "image:<acs.version>-<build-number>" on Quay -->
             <properties>
-                <image.tag>${acs.version}-${build-number}</image.tag>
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -195,7 +195,7 @@ jobs:
     - name: "Update latest and Single Pipeline <acs>-<build> images"
       stage: docker_latest
       # Remove the pipeline profile to avoid sending an image tagged with the build number to the single pipeline
-      script: travis_retry travis_wait 30 mvn -B -V clean install -DskipTests -Dmaven.javadoc.skip=true -Dbuild-number=${TRAVIS_BUILD_NUMBER} -Pinternal,pipeline
+      script: travis_retry travis_wait 30 mvn -B -V clean install -DskipTests -Dmaven.javadoc.skip=true -Dbuild-number=${TRAVIS_BUILD_NUMBER} -Dbranch=${TRAVIS_BRANCH} -Pinternal,pipeline
```
mteodori, line 198: should this be acs.branch? or just branch in the pom.xml?

tests/pipeline-all-amps/pom.xml
```diff
@@ -61,7 +61,7 @@
             <id>internal</id>
             <!-- publishes "image:<acs.version>-<build-number>" on Quay -->
             <properties>
-                <image.tag>${acs.version}-${build-number}</image.tag>
+                <image.tag>${acs.branch}-${build-number}</image.tag>
```
mteodori, line 64: @alandavis can this affect other non master HEAD branch builds we're running?
using the slash in the branch name would create an invalid image tag and probably break the build

eromano: I think we do it only for develop

alandavis: You are correct. We have to support other ACS releases. This change would switch from tags names such as

6.2.3-9876
6.2.2-4567
7.0.0-1234

To 
release/6.2.N-9876     -- note the N and /
release/6.2.2-4567
master-1234

I really don't have a view as to what the tag names are for the single pipeline, other than that I would have though what we have at the comment would have been simpler, but I'm not consuming the images.

mteodori: is it correct this job runs on master and release/* or can we just do master? that would solve the problem

https://github.com/Alfresco/acs-packaging/blob/master/.travis.yml#L38-L39
https://github.com/Alfresco/acs-packaging/blob/master/.travis.yml#L195-L198

alandavis: It runs currently on both.
If your regex is: 7\.0\.0-[0-9]+
is there a problem?

I don't know what problem you are trying to solve.

alandavis:
Closing for now. We do need to support different ACS versions, but it is not clear we also need further branches.

PR #1862, opened 2021-02-05
OPSEXP-718: Use DTAS for smoke testing the pipeline-all-amps-repo image

gavincornwell:
- Removed Centera AMP as it's not used in the pipeline
- Fix S3 connector config & use custom admin username but default password
- Fixed startup script (was passing even though system was failing to start)
- Run DTAS after system is running

NOTE: The AGS AMP does not report the same version as the artefact version so it had to be hardcoded in DTAS config. Same for ACS version, but that shouldn't change as often.


alandavis (approved):
LGTM

PR #1867, opened 2021-02-12
SEARCH-2688 Elasticsearch admin console tests.

tpage-alfresco:
See also https://github.com/Alfresco/alfresco-enterprise-repo/pull/255


pom.xml
```diff
@@ -10,12 +10,12 @@
         <groupId>org.alfresco</groupId>
         <artifactId>alfresco-enterprise-repo</artifactId>
         <relativePath>../alfresco-enterprise-repo/pom.xml</relativePath>
-        <version>8.423</version>
+        <version>8.422-SNAPSHOT</version>
```
jesty, line 13: can you remove those version changes? Thanks

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchAdminConsoleTests.java
```diff
@@ -0,0 +1,151 @@
...
+        Step.STEP("Refresh the admin console.");
+        document = loadSearchAdminPage();
+        Step.STEP("Get the new number of indexed and indexable documents.");
+        int newIndexedDocs = Integer.valueOf(document.select("#elasticsearchDocumentCount span.value").text());
+        int newIndexableDocs = Integer.valueOf(document.select("#repositoryNodesCount span.value").text());
+
+        // Check the numbers have increased (in theory by one each, but there may be other documents being indexed too.
+        Assert.assertTrue(newIndexedDocs > indexedDocs, "Expected the number of indexed documents to increase, was "
+                + indexedDocs + ", now " + newIndexedDocs);
+        Assert.assertTrue(newIndexableDocs > indexableDocs, "Expected the number of indexable documents to increase, was "
```
jesty, line 111: because we are indexing only one document, can we verify that `newIndexableDocs = indexableDocs+1`?

jesty: I noticed the comment, why there are other documents being indexed too?

jesty (changes requested):

pom.xml
tpage-alfresco: Oh - good reminder. I have to include them so that the build passes, but once the feature branch is finalised then I need to revert this before merging.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchAdminConsoleTests.java
tpage-alfresco: There weren't any other documents being indexed in my local testing (and checking `y == x + 1` worked), but I wasn't sure that we could guarantee this would always be the case. I'm happy to change it for now, and we could change it back if we see intermittent failures?

tests/tas-elasticsearch/pom.xml
```diff
@@ -49,6 +49,11 @@
             <artifactId>spring-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jsoup</groupId>
+            <artifactId>jsoup</artifactId>
+            <version>1.13.1</version>
```
tpage-alfresco, line 55: This should be a test dependency.

jesty (approved):
LGTM

PR #1895, opened 2021-03-03
SEARCH-2707 add a test to verify that the system is able to index and search for dates before 1970

jesty:
Because values for milliseconds-since-the-epoch and seconds-since-the-epoch must be non-negative we have to test what happens when we use a negative epoch when we create a node.

This PR should be green when https://github.com/Alfresco/acs-event-model/pull/12 will be merged.


robfrank (approved):
LGTM

PR #2006, opened 2021-06-28
SEARCH-2952 Fix reindexing tests

tpage-alfresco:
Update ESC version to 3.0.0-A4.  Change queries to avoid exact term search limitations.  Add environment variables to allow ESC components to connect to AMQ and the transform service.


jesty (approved):
LGTM

PR #2051, opened 2021-08-05
[AGS] Publish AGS-specific docker images from acs-packaging

elizastan:


CezarLeahu (commented):
These two modules must be moved into an `<id>ags<id>` maven profile.

CezarLeahu (approved):

PR #2052, opened 2021-08-06
SEARCH-2976 Avoid overwriting "latest" Docker Images for "release" br…

aborroy:
…anches when pushing images for pipeline


killerboot (approved):
Do you need to change the regexp in terraform for one pipeline?

killerboot (commented):
Another question. Do these docker images have an expiry date? It needs to be added so that quay can clean up automatically

aborroy:
This approach is not valid, as this Travis Job is used to push pipeline images but also to push Repo and Share images.

PR #2053, opened 2021-08-06
SEARCH-2976 Avoid overwriting "latest" Docker Images for "release" br…

aborroy:
…anches when pushing images for pipeline

Expiration Date is not an option for "docker-maven-plugin"


killerboot (approved):
It is not ideal to have a vendor specific info in the docker image, but I think it is ok to proceed with that without writing additional scripts.

PR #2061, opened 2021-09-01
SEARCH-3010 E2E test for path reindexing.

tpage-alfresco:


tpage-alfresco:
Rebased the changes onto the latest acs-packaging as this should now be able to use the path query work from master of alfresco-enterprise-repo.

PR #2063, opened 2021-09-02
Feature/search 3020 reindexing services disabling

aborroy:


tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchReindexingTests.java
```diff
@@ -150,6 +150,7 @@ public void testReindexerFixesBrokenIndex() throws Exception
 
         // TIDY
         // Restart ElasticsearchConnector.
+        cleanUpIndex();
```
tpage-alfresco, line 153: This doesn't seem like a good idea. If a test needs there to be no documents in the index then it should clean the index at the start, rather than adding this to the end of every other test.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchReindexingTests.java
```diff
@@ -178,8 +179,112 @@ public void testRecreateIndex() throws Exception
...
+    private void internalTestEnabledFeatures(
+        Boolean metadataIndexingEnabled,
+        Boolean contentIndexingEnabled,
+        Boolean pathIndexingEnabled,
+        String queryString,
+        Boolean expectingDocNameAsResult
+    ) throws Exception
+    {
+
+        // Initial timestamp for reindexing by date: this will save reindexing time for these tests
```
tpage-alfresco, line 204: Does it actually save a measurable reindexing time?  If it does then there are probably other tests that could benefit from this, if not then IMO we should simplify by removing it (since we can reindex everything by just supplying `reindexByDate`).

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchReindexingTests.java
```diff
@@ -178,8 +179,112 @@ public void testRecreateIndex() throws Exception
...
+        ZonedDateTime now = ZonedDateTime.now(Clock.systemUTC());
+        String testStart = DateTimeFormatter.ofPattern("yyyyMMddHHmm").format(now.minusMinutes(1));
+
+        // GIVEN
+        // Stop ElasticsearchConnector
+        AlfrescoStackInitializer.liveIndexer.stop();
+        // Create document
+        String documentName = createDocument();
+        // Delete index documents
+        cleanUpIndex();
```
tpage-alfresco, line 214: Do we need to do this?

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchReindexingTests.java
```diff
@@ -178,8 +179,112 @@ public void testRecreateIndex() throws Exception
...
+            "cm:name:'<DOCUMENT_NAME>' AND TEXT:'content'", false);
+    }
+
+    @Test(groups = TestGroup.SEARCH)
+    public void testRecreateIndexWithNoMetadataAndContent() throws Exception
+    {
+        // When not using metadata, document shouldn't be present in Elasticsearch index,
+        // since metadata reindexing process is indexing also permissions
+        internalTestEnabledFeatures(false, true, false,
+            "cm:name:'<DOCUMENT_NAME>'", false);
```
tpage-alfresco, line 263: Do we know that this query won't hit the DB?

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchReindexingTests.java
```diff
@@ -178,8 +179,112 @@ public void testRecreateIndex() throws Exception
...
+        internalTestEnabledFeatures(true, false, true,
+            "cm:name:'<DOCUMENT_NAME>' AND PATH:'/app:company_home/st:sites/cm:" + testSite + "/cm:documentLibrary/cm:<DOCUMENT_NAME>'", true);
+    }
+
+    @Test(groups = TestGroup.SEARCH)
+    public void testRecreateIndexWithMetadataAndContentAndPath() throws Exception
+    {
+        internalTestEnabledFeatures(true, true, true,
+            "cm:name:'<DOCUMENT_NAME>' AND TEXT:'content' " +
+                "AND PATH:'/app:company_home/st:sites/cm:" + testSite + "/cm:documentLibrary/cm:<DOCUMENT_NAME>'", true);
```
tpage-alfresco, line 278: Great that this is working now! When I tried before then I couldn't get absolute paths to work.

tpage-alfresco:
Really nice coverage :+1: Mostly some questions about whether we can delete some steps from the test or not.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchReindexingTests.java
aborroy: It's saving 2-3 seconds per test... Is that worthy?

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchReindexingTests.java
aborroy: Solved

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchReindexingTests.java
aborroy: Is it better to create a single document for every tests?
This approach is creating a new document for each test and cleaning the existing Elasticsearch index to be sure that only the new document is target of the reindexing process.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchReindexingTests.java
aborroy: We need to add `AND cm:name:*` to avoid that. Thanks!

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchReindexingTests.java
tpage-alfresco: That sounds like a really good saving!  We should (perhaps as a separate PR) make the same change to the other tests if we can.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchReindexingTests.java
tpage-alfresco: But do we mind if the reindexing process affects more nodes?

tpage-alfresco (approved):
LGTM!

PR #2064, opened 2021-09-06
ATS-955 Upgrade AIS, S3-Connector, Azure-Connector

PunchedChimera:
Upgrade the following:

AIS -> 1.4.0-A1
S3-Connector -> 4.1.0-A2
Azure-Connector -> 2.0.1


alandavis (approved):
LGTM - Yes you need the new library versions with the S3 and Azure Connectors

PR #2066, opened 2021-09-06
SEARCH-3010 E2E tests for live indexing of paths.

tpage-alfresco:


agazzarini:
Hi @tpage-alfresco the approach sounds good. A couple of things

- re-reading the [PATH Queries Design](https://alfresco.atlassian.net/wiki/spaces/TECH/pages/261718775/Path+Queries+Support) it seems full URIs (in namespaces) are not supported 
- what about a test for each scenario in the table reported in SEARCH-2790? I would include also "Disabled" tests for those things that don't work (the 1st row)

agazzarini (approved):

PR #2067, opened 2021-09-07
Refactor more tests to use SearchQueryService.

tpage-alfresco:
This simplifies several tests and also improves some assertion messages which should help us to debug intermittent issues.


agazzarini (approved):
LGTM @tpage-alfresco !

PR #2073, opened 2021-09-29
Create an acs-all-amps image that contains a keystore for saml

mauriziovitale:


gicappa (approved):
LGTM!

andrea-ligios (approved):
LGTM (we could even generate the JKS on the fly in a RUN command in the Dockerfile, but this is definitely enough for now)

Nope, we cannot generate it on the fly because otherwise, the key stored in the keycloak client won't be the same.

jamalkm (approved):

mauriziovitale:
> Nope, we cannot generate it on the fly because otherwise, the key stored in the keycloak client won't be the same.


Nope, we cannot generate it on the fly because otherwise, the key stored in the keycloak client won't be the same.

PR #2082, opened 2021-10-13
SEARCH 2028 Adding e2e test for updating folder path

klalia:


klalia:
LGTM, sorry for the additional commits, I messed up the base branch. Please double check again before merging @robfrank just in case but the build is back to green and the tests are running :)

PR #2103, opened 2021-11-11
ACS-2217 add test on root node without wildcards

robfrank:


scripts/travis/build.sh
```diff
@@ -97,7 +97,7 @@ else
 fi
 
 # Build the current project
-mvn -B -V -q install -DskipTests -Dmaven.javadoc.skip=true -Pbuild-docker-images -Pags ${REPO_IMAGE} ${SHARE_IMAGE}
+mvn -B -V install -DskipTests -Dmaven.javadoc.skip=true -Pbuild-docker-images -Pags ${REPO_IMAGE} ${SHARE_IMAGE}
```
tpage-alfresco, line 100: Is this change desired?  I can see it's useful for debugging, but does it bloat the log?  Previously we've had success by enabling this, but turning off the logging around "Downloading..." and "Downloaded..." messages - perhaps that would be a compromise?

pom.xml
```diff
@@ -10,12 +10,12 @@
...
         <relativePath>../alfresco-enterprise-repo/pom.xml</relativePath>
-        <version>14.26</version>
+        <version>14.27-SNAPSHOT</version>
     </parent>
 
     <properties>
-        <dependency.alfresco-enterprise-repo.version>14.26</dependency.alfresco-enterprise-repo.version>
-        <dependency.alfresco-enterprise-share.version>14.17</dependency.alfresco-enterprise-share.version>
+        <dependency.alfresco-enterprise-repo.version>14.27-SNAPSHOT</dependency.alfresco-enterprise-repo.version>
+        <dependency.alfresco-enterprise-share.version>14.18-SNAPSHOT</dependency.alfresco-enterprise-share.version>
```
tpage-alfresco, line 18: These all need changing back to release versions before this can be merged.

tpage-alfresco:
The test looks good. There are a couple of things to look at before merging.

pom.xml
robfrank: yes, I should rebase before merging: at that point, the build will be red (I guess): these versions are set to that values to be bound to enterprise repo branch

scripts/travis/build.sh
robfrank: was for debugging, will be removed  by rebase

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchPathIndexingTests.java
```diff
@@ -135,6 +135,13 @@ public void testRootNodes()
         searchQueryService.expectResultsFromQuery(query, testUser, "categories", "Company Home");
     }
 
+    @Test(groups = TestGroup.SEARCH)
+    public void testRootNodesWithoutWildcard()
+    {
+        SearchRequest query = req("PATH:\"/\" AND name:*");
+        searchQueryService.expectResultsFromQuery(query, testUser, "categories", "Company Home");
```
tpage-alfresco, line 142: I don't think this is right. Only the root node should have the path `/`.

tpage-alfresco (approved):
LGTM!

PR #2111, opened 2021-11-15
ACS-2172 Bump S3-connector-4.1.1

PunchedChimera:


.travis.yml
```diff
@@ -297,9 +297,9 @@ jobs:
           on:
             all_branches: true
         - provider: script
-            script: bash scripts/travis/copy_share_image_to_dockerHub.sh
-            on:
-              all_branches: true
+          script: bash scripts/travis/copy_share_image_to_dockerHub.sh
```
montgolfiere, line 300: @PunchedChimera Was this an intentional change (formatting only) ?

montgolfiere (approved):

.travis.yml
PunchedChimera: Yes, the incorrect indentation actually caused the travis.yml to not be parsable by the build agents. This fixes it.

PR #2124, opened 2021-11-24
ACS-2171 use package reference in test suite instead of classes

robfrank:
use package reference in test suite instead of classes
refactor tests: move ES basicAuth test to dedicated package


klalia:
LGTM :)

klalia (approved):

PR #2140, opened 2022-01-13
ACS-2441 - adjust log levels for TAS-based ES tests

robfrank:
log4j: log level to debug
testng: verbose to 1


robfrank:
After taking a look at other tests suites, I think these log settings are better and we are not losing details in case of test failures. 
Keeping it so verbose is risky, we can run out of log space just adding a new test, that's what happened on the failing build: we ran out of log space due to some lines of maven logging.
Moreover, setting verbose level on testing suites, enable to have a  detailed view of executed tests

PR #2147, opened 2022-01-14
Update CHANGELOG.md to include 7.1.1 changes

SaraAspery:


tpage-alfresco (approved):
The changes look fine (I didn't really check them though).  I think this looks like it will require a lot of manual effort to maintain and trying to create this as part of the release process seems burdensome.  Please can we review why we're maintaining this file here, rather than having an automatic report from Jira?

isandhu (commented):
There is a mis-mtach with what we added to the release notes, which was based on this [query](https://alfresco.atlassian.net/issues/?jql=project%20%3D%20MNT%20and%20fixVersion%20%3D%20%22ACS%207.1.1%22%20AND%20resolution%20in%20(Fixed%2C%20Done)) that matches closely a Confluence [page](https://alfresco.atlassian.net/wiki/spaces/TECH/pages/917865028/ACS+7.1.1+Mx+Release+Notes) that Sara had started.  I don't really want to cross-check it all manually at this stage.

alandavis (approved):
Worth merging back to the release/7.1.1 branch even if we don't create a HF.

alandavis:
We are no longer maintaining CHANGELOG and has been removed from the latest versions.  The PMs are now responsible for creating release notes, and will have used the text provided here.

isandhu:
Acknowledged.  Will let PM know that this will no longer be maintained. Thanks.

PR #2169, opened 2022-02-04
ACS-2290 Validate the ACS version properties in community repo.

tpage-alfresco:


scripts/travis/build.sh
```diff
@@ -65,6 +65,17 @@ else
   pullUpstreamTag "${COM_UPSTREAM_REPO}" "${COM_DEPENDENCY_VERSION}"
 fi
```
alandavis, line 67: I'm not sure this will work with HFs where there is a label too (generally edited only in the enterprise repo). You would need to see what the CX team do with their WoW. 

This check would only really come into its own when creating new branches. I don't think I have ever forgotten to change the acs-packaging version.

tpage-alfresco: The check will work fine with labels or HF branches. It just checks that the acs-packaging version starts with `[major].[minor].[revision]` where those variables come from alfresco-community-repo. You're right that the check is only useful when updating those three digits of the release.

tpage-alfresco:
@jottley @evasques @vrmoreira Adding you guys to this review to check that the test won't break anything for HF branches. Currently it's just against master, but I plan to cherry-pick it back to all SP branches.

montgolfiere (approved):

alandavis (approved):

DenysVuika (approved):

tpage-alfresco:
Talking with @evasques this PR seems ok but will not cover the hotfix version which comes from enterprise repo. I'll try to create a separate PR to validate this.

PR #2174, opened 2022-02-15
ACS-2290 Improve version validation to include hotfix version too.

tpage-alfresco:


evasques (approved):
LGTM

alandavis:
I don't think having these check or those in the previous PR is a good idea. They force you to modify the upstream projects even if all you want to do is modify one of the packaging projects and create a new release by setting the RELEASE_VERSION in the .travis.yml file. Yes I accept the version.properties file will have the wrong value, but sometimes that is the better option.

evasques:
@alandavis just to clarify a bit: the validation is on the 'version line' (as in 7.1.0) not on the exact internal dependency versions. For example: if we want to release on master with mixed versions of community-repo (CR) and enterprise-repo(ER): 14.108 CR with 14.104 ER (that don't 'match') the validation Tom added just checks if 14.108 CR is on 7.2.0 and 14.104 ER is also on 7.2.0, same as the ACS-Packaging branch you are on.

This validation however will prevent from building on (for example) acs-packaging master mixing 14.108CR and 13.2ER as 13.2ER version line is 7.1.2

We could also be doing this by verifying if we are building (using the example above) using a 14.X CR with a 14.X ER (as Tom implemented for the Share version validation and we can switch to if we eventually move to a date based version line)

alandavis:
I have probably got the wrong end of the stick. I now see the checks don't look at the `acs.version.label`.

What I really wanted to avoid was having to update that value in CR or ER when we are under time pressures. There have been times when we just need to do another release by setting the 'RELEASE_VERSION' to the next HF version. e.g. the release build failed half way through with a network timeout. Having to go back to CR or ER and do all the builds again would be a right pain and add at least half a day.

RE: mismatched versions
It is not the versions of CR, ER and ES poms, but the acs.version.major, minor, revision and label which get used in version.properties file.

RE: Hyland version numbering
1 pint beer bet that PMs insist that the artifacts contain the string YY.X at some point. We might just want to also change the CR, ER, ES pom versions to match. Fortunately 22 is > the value used on master.

tpage-alfresco:
Ok - I'm going to give this a go and we can see how we get on.

PR #2176, opened 2022-02-18
ACS-2576 Add possibility to override the es connector tag

pzhyland:


tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/AlfrescoStackInitializer.java
```diff
@@ -110,6 +111,16 @@ public void initialize(ConfigurableApplicationContext configurableApplicationCon
 
     }
 
+    public static String getElsticsearchConnectorImageTag()
```
tpage-alfresco, line 114: Missing "a" in Elastic.

tests/environment/.env
```diff
@@ -7,3 +7,6 @@ ACTIVEMQ_TAG=5.16.1
 AIMS_TAG=1.2
 SYNC_SERVICE_TAG=3.5.0
 ES_TAG=7.10.1
+#By default the connector tag is set to the value of `dependency.elasticsearch-shared.version` maven property.
+#Following env variable allow you to override this value. It is useful for testing.
```
tpage-alfresco, line 11: Please could you add "It should be left blank" or something like that so that it's obvious that a PR shouldn't change this value.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/AlfrescoStackInitializer.java
pzhyland: Thanks!. Done.

tests/environment/.env
pzhyland: Done.

kcichonczyk (approved):

PR #2179, opened 2022-02-18
ACS-2290 Validate versions for 7.1.1.

tpage-alfresco:


tpage-alfresco:
I think the right way to do this is to automatically update the versions rather than validating them.

PR #2180, opened 2022-02-18
ACS-2290 Validate versions for 7.1.0.

tpage-alfresco:


tpage-alfresco:
> Referenced version of community repo specifies "7.1.0" in pom.xml and enterprise repo specifies ".3", but this is 7.1.0.4-SNAPSHOT.

tpage-alfresco:
I think the right way to do this is to automatically update the versions rather than validating them.

PR #2196, opened 2022-03-17
MNT-22864 Publish all maven artifacts to enterprise-releases.

tpage-alfresco:


mpichura (approved):
LGTM and I hope rebuilding with all TAS tests in place will not increase failure rate in travis (any foreseeable test failures lurking?).
I wonder if it would be possible to omit all-tas-tests profile here (if it is run in any previous stages)?

tpage-alfresco:
Yes - I think we can skip all the tests actually.

tpage-alfresco:
Just to add that the all-tas-tests profile by itself doesn't cause the tests to be run, but simply includes the submodule so that it gets published.

krdabrowski (approved):

montgolfiere (approved):

PR #2198, opened 2022-03-22
Build should not go red after each release - a test too far

alandavis:
Test ensures the build will go red after each HF is released and also stops us from just using the next number in sequence (which we have done several times) without going back to change a setting in enterprise repo and waiting several hours for builds.


alandavis:
Note: The PR build was red, but I have still merged as the build has shown up another issue. https://alfresco.atlassian.net/browse/ACS-2734

PR #2202, opened 2022-04-06
ACS-2802 CMIS suffix query test.

tpage-alfresco:


tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -31,12 +31,13 @@
 public class ElasticsearchCMISTests extends AbstractTestNGSpringContextTests
 {
     private static final String PREFIX = getAlphabeticUUID() + "_";
+    private static final String SUFFIX = "_" + getAlphabeticUUID();
     private static final String UNIQUE_WORD = getAlphabeticUUID();
-    private static final String FILE_0_NAME = PREFIX + "test.txt";
-    private static final String FILE_1_NAME = PREFIX + "another.txt";
-    private static final String FILE_2_NAME = PREFIX + "user1Old.txt";
+    private static final String FILE_0_NAME = PREFIX + "test.txt" + SUFFIX;
```
montgolfiere, line 36: Does it make more sense (from user's point of view) to have the suffix before the period, eg.

test_suffix.txt

Or else maybe test both ways ?

tpage-alfresco: I did some experimentation with this against ES. There's a bug that a prefix is only matched if there's nothing else in the term after it. Weirdly `PREFIX%` matches `backwardsPREFIX` but not `PREFIXbody`. To work around this issue for now I've made sure that all the test files have underscores separating the prefix from the rest of the filename.

gevagel (approved):

PR #2205, opened 2022-04-11
ACS-2803: Add test to check if IN syntax works against ES

gevagel:


montgolfiere:
I see this failed on Travis with 500 error. Anything obvious in logs (eg. if not on Travis then when run locally) otherwise needs to be debugged.

```
[ERROR] checkInSyntax(org.alfresco.elasticsearch.ElasticsearchCMISTests)  Time elapsed: 20.619 s  <<< FAILURE!
java.lang.AssertionError: Maximum retry period reached, test failed.
	at org.alfresco.elasticsearch.ElasticsearchCMISTests.checkInSyntax(ElasticsearchCMISTests.java:155)
Caused by: java.lang.AssertionError: Status code is not as expected. expected [200] but found [500]
	at org.alfresco.elasticsearch.ElasticsearchCMISTests.checkInSyntax(ElasticsearchCMISTests.java:155)
```

montgolfiere:
@gevagel Have you managed to debug where (& why) this fails ?

montgolfiere:
@gevagel What's the plan for this one ?

montgolfiere:
Thanks @gevagel  - good to know.

Are you sure you really need custom content model ?

AFAIK, the tests via TestContainers allows for programmatic control of running specific tests with specific "feature flags" (test env config properties) such as with & without exact term matching, cc @tpage-alfresco 

- https://docs.alfresco.com/search-enterprise/latest/admin/#pre-indexing-considerations
- https://docs.alfresco.com/search-enterprise/latest/using/#search-in-fields

montgolfiere:
See also: https://github.com/Alfresco/alfresco-enterprise-repo/pull/773

It would be good to review and approve both together. At minimum the happy path (ie. exact term enabled).

It might be that here we also go one step further (follow-on PR ?) and test with and without exact term (ie. -ve test) (eg. using current programmatic TestContainers approach).

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -147,6 +147,14 @@ public void checkPermissionForUser2()
         searchQueryService.expectResultsFromQuery(query, user2, FILE_2_NAME, USER_2_FILE_NAME);
     }
 
+    @TestRail (description = "Check IN('value1','value2') syntax works.", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
+    @Test (groups = TestGroup.SEARCH)
+    public void checkInSyntax()
+    {
+        SearchRequest query = req("cmis", "SELECT * FROM cmis:document WHERE cmis:name IN ('" + FILE_0_NAME + "') AND CONTAINS('*')");
```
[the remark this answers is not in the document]
montgolfiere: Make this consistent with https://github.com/Alfresco/alfresco-enterprise-repo/pull/773/files#diff-2466e672e08f8f55c9f5273d5f2a5555f148ce057aaf137da65bbd7df917f35cR57 (ie. check at least two IN values).

montgolfiere (changes requested):
Added some initial comments. 

We should make this at least consistent with similar ent-repo test, ie. minimum, test with exact match enabled).

Also consider additional test scenarios (given separate test config, maybe separate test class) with alternative (arguably negative) behaviour, ie. exact match not enabled.

tests/tas-elasticsearch/src/test/resources/exactTermSearch.properties
```diff
@@ -0,0 +1,7 @@
+# Exact Term search is switched off by default as it introduces index size overhead.
+#to enable it, please uncomment the following on a datatype or property name basis
+#alfresco.cross.locale.datatype.0={http://www.alfresco.org/model/dictionary/1.0}text
+#alfresco.cross.locale.datatype.1={http://www.alfresco.org/model/dictionary/1.0}content
+#alfresco.cross.locale.datatype.2={http://www.alfresco.org/model/dictionary/1.0}mltext
+#alfresco.cross.locale.property.0={http://www.alfresco.org/model/content/1.0}content
+alfresco.cross.locale.property.1={http://www.alfresco.org/model/content/1.0}name
```
montgolfiere, line 7: Should this by `.0` (or does it not matter if only one property) ?

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -147,6 +147,22 @@ public void checkPermissionForUser2()
         searchQueryService.expectResultsFromQuery(query, user2, FILE_2_NAME, USER_2_FILE_NAME);
     }
 
+    @TestRail (description = "Check that exact term search works. Needs exact term search to be enabled to pass.", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
+    @Test (groups = TestGroup.SEARCH)
+    public void matchDocumentName()
+    {
+        SearchRequest query = req("cmis", "SELECT * FROM cmis:document WHERE cmis:name = '" + FILE_0_NAME + "' AND CONTAINS('*')");
```
montgolfiere, line 154: Do we really want the CONTAIN('*') here ? 

Wouldn't it be better to disable the option to go via DB (TMDQ) when starting repo ? At least for Elasticsearch CMIS Tests (and possibly AFTS ones as well) ?

Note: for now it requires the following property name (until we look to remove "solr." part):

`-Dsolr.query.cmis.queryConsistency=EVENTUAL`

And remove &/or refactor any tests that do not yet work, eg. basicCMISQueryAgainstDB(). See for example:

https://github.com/Alfresco/alfresco-enterprise-repo/pull/809/files#diff-2466e672e08f8f55c9f5273d5f2a5555f148ce057aaf137da65bbd7df917f35cR41-L44

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -147,6 +147,22 @@ public void checkPermissionForUser2()
...
+        searchQueryService.expectResultsFromQuery(query, user1, FILE_0_NAME);
+    }
+
+    @TestRail (description = "Check IN('value1','value2') syntax works. Needs exact term search to be enabled to pass.", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
+    @Test (groups = TestGroup.SEARCH)
+    public void checkInSyntax()
+    {
+        SearchRequest query = req("cmis", "SELECT * FROM cmis:document WHERE cmis:name IN ('" + FILE_0_NAME + "', '" + FILE_1_NAME + "') AND CONTAINS('*')");
```
montgolfiere, line 162: See comment above, re: CONTAINS('*')

montgolfiere (commented):
Thanks @gevagel and @KikoUK 

I've added a few minor comments & possible improvements. 

For now, I suggest we try to be aligned and consistent with BasicQueryIT (in alfresco-enterprise-repo) ... at least where it makes sense.

KikoUK (approved):
LGTM subject to ✅ build and Jan's comments being addressed.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -147,6 +147,22 @@ public void checkPermissionForUser2()
         searchQueryService.expectResultsFromQuery(query, user2, FILE_2_NAME, USER_2_FILE_NAME);
     }
 
+    @TestRail (description = "Check that exact term search works. Needs exact term search to be enabled to pass.", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
```
tpage-alfresco, line 150: I think we're really checking that we can match a document name (as the method name says). This uses exact term search, but I don't think an end user would think of it that way.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
tpage-alfresco: This sounds like a great improvement to these tests, but let's do it as a separate ticket.

montgolfiere: Hopefully just a couple of lines (assuming no other fallout). But yes, it could also be done as part of ACS-2920 (if not sooner).

tests/tas-elasticsearch/src/test/resources/exactTermSearch.properties
```diff
@@ -0,0 +1,7 @@
+# Exact Term search is switched off by default as it introduces index size overhead.
+#to enable it, please uncomment the following on a datatype or property name basis
+#alfresco.cross.locale.datatype.0={http://www.alfresco.org/model/dictionary/1.0}text
+#alfresco.cross.locale.datatype.1={http://www.alfresco.org/model/dictionary/1.0}content
+#alfresco.cross.locale.datatype.2={http://www.alfresco.org/model/dictionary/1.0}mltext
+#alfresco.cross.locale.property.0={http://www.alfresco.org/model/content/1.0}content
```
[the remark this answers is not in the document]
montgolfiere: As per @tpage-alfresco we could remove commented lines.

Instead, we could also add a comment line which cross-references to similar test config for embedded IT tests running in alfresco-enterprise-repo.

https://github.com/Alfresco/alfresco-enterprise-repo/blob/master/repository/src/test/resources/alfresco/search/elasticsearch/config/exactTermSearch.properties

Interestingly that example has cm:content but not cm:name. Instead, it has d:text and d:mltext (presumably to cover all indexed text properties) ?

montgolfiere (approved):
See @tpage-alfresco comments.

Otherwise, I'm ok if we merge and then plan to improve and extend as part of additional sub-tasks, including fix for ACS-2920.

PR #2206, opened 2022-04-12
ACS-2798: Add tests to support "SELECT field" syntax

KikoUK:
- Add tests for SELECT field and return a list

### DIFF
--- a/tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
+++ b/tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
@@ -104,6 +104,22 @@ public class ElasticsearchCMISTests extends AbstractTestNGSpringContextTests
         searchQueryService.expectSomeResultsFromQuery(query, user1);
     }
 
+    @TestRail (description = "Verify that we can return a list of fields from a document via a CMIS query against Elasticsearch.", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
+    @Test (groups = TestGroup.SEARCH)
+    public void returnFieldListFromDoc()
+    {
+        SearchRequest query = req("cmis", "SELECT cmis:name, cmis:objectId, cmis:lastModifiedBy, cmis:creationDate, cmis:contentStreamFileName FROM cmis:document");
+        searchQueryService.expectSomeResultsFromQuery(query, user1);
+    }
+
+    @TestRail (description = "Verify that we can return a list of fields from a folder via a CMIS query against Elasticsearch.", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
+    @Test (groups = TestGroup.SEARCH)
+    public void returnFieldListFromFolder()
+    {
+        SearchRequest query = req("cmis", "SELECT cmis:name, cmis:parentId, cmis:path, cmis:allowedChildObjectTypeIds FROM cmis:folder");
+        searchQueryService.expectSomeResultsFromQuery(query, user1);
+    }
+
     @TestRail(description = "Verify that we can perform a basic CMIS query against the DB.", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
     @Test(groups = TestGroup.SEARCH)
     public void basicCMISQueryAgainstDB()


tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -104,6 +104,22 @@ public void basicCMISQuery()
         searchQueryService.expectSomeResultsFromQuery(query, user1);
     }
 
+    @TestRail (description = "Verify that we can return a list of fields from a document via a CMIS query against Elasticsearch.", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
+    @Test (groups = TestGroup.SEARCH)
+    public void returnFieldListFromDoc()
+    {
+        SearchRequest query = req("cmis", "SELECT cmis:name, cmis:objectId, cmis:lastModifiedBy, cmis:creationDate, cmis:contentStreamFileName FROM cmis:document WHERE CONTAINS('*')");
```
montgolfiere, line 111: Does the list of field name affect the results ?

If yes, can we check ?

If no, then i assume there must be a negative test - eg. fields that don't exist and cause semantic error ?

montgolfiere, line 111: Does the list of field name affect the results ?

If yes, can we check ?

If no, then i assume there must be a negative test - eg. fields that don't exist and cause semantic error ?

montgolfiere: OK, but do the response differ with different valid fields ?

KikoUK: We tried that with @gevagel in enterprise repo and seems that it always returns the same result from the SELECT query regardless of the requested field/fields so the decision was the wait for @tpage-alfresco to confirm that SELECT field is supported

montgolfiere: FYI: I've tried CMIS Workbench client against a Repo with Solr (SSv2), eg.

`SELECT cmis:name FROM cmis:document WHERE CONTAINS('email')`

I notice client only shows one column (compared to SELECT * which shows multiple columns). I have not analyzed network traffic to see if this is done client-side or server-side (when querying via Solr)

montgolfiere: Note: with Solr, I think I can see 400 in proxy as expected, eg.

`proxy_1                      | 86.15.70.161 - admin [13/Apr/2022:16:36:02 +0000] "POST /alfresco/api/-default-/public/cmis/versions/1.1/browser HTTP/1.1" 400 96 "-" "Apache-Chemistry-OpenCMIS/1.1.0 (Java 11.0.10; Mac OS X 12.3.1)"`

On client-side (using CMIS Workbench), I see:

```
CMIS Exception: invalidArgument Invalid column for cmis:document cmis:nam
  Error Content: {"exception":"invalidArgument","message":"Invalid column for cmis:document cmis:nam"}
```

montgolfiere (changes requested):
I think we need -ve tests (either here in e2e or else IT) for invalid field name

=> note: ideally I'd expect 400 but apparently with Solr it is 500 :-(

Also, confirm behaviour (of response) when selecting valid field name(s) compared to CMIS Query via Solr.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
KikoUK: Can you try it against Elasticsearch?

montgolfiere: Unfortunately not - see: https://alfresco.atlassian.net/browse/ACS-2860

montgolfiere:
Why did we close ?

KikoUK:
> 

Ticket was closed by @tpage-alfresco and he mentioned that tests are not needed.

PR #2218, opened 2022-05-10
ACS-2864 Script the creation of new repo, share and packaging branches after an ACS release

alandavis:
Script to creates HotFix and ServicePack (.N) branches for repo, share and packaging projects, and updates the master or ServicePack branches ready for the next release.

- This script may be used following releases >= 7.2.1.
- Master must have been used for major or mion releases and .N branches for service pack releases.

https://alfresco.atlassian.net/wiki/spaces/TECH/pages/962298737/ACS+Release+Process+-+Creating+the+branches+after+a+release

This script will be cherry picked back to release/7.2.N

### DIFF
--- a/scripts/dev/newReleaseBranch.sh
+++ b/scripts/dev/newReleaseBranch.sh
@@ -202,10 +202,56 @@ getCurrentProject() {
   basename ${pwd}
 }
 
-setVersion() {
+setAcsVersionLabelInEnterpriseRepoHotFixes() {
+  local version="${1}"
+  local branchType="${2}"
+
+  if [[ `getCurrentProject` == "alfresco-enterprise-repo" && "${branchType}" == "HotFix" ]]
+  then
+    echo "${prefix}    set <acs.version.label>"
+    ed -s pom.xml &>${loggingOut} << EOF
+/.*acs.version.label.*$/s//        <acs.version.label>.1<\/acs.version.label> <!-- ${version}.<acs.version.label> -->/
+wq
+EOF
+  fi
+}
+
+setNextReleaseVersion() {
   local version="${1}"
 
-  echo "${prefix}    TODO setVersion ${version}"
+  echo "${prefix}    set - RELEASE_VERSION=${version}"
+  ed -s .travis.yml &>${loggingOut} << EOF
+/.*- RELEASE_VERSION.*$/s//    - RELEASE_VERSION=${version}/
+wq
+EOF
+}
+
+setNextDevelopmentVersion() {
+  local version="${1}"
+
+  echo "${prefix}    set - DEVELOPMENT_VERSION=${version}"
+  ed -s .travis.yml &>${loggingOut} << EOF
+/.*- DEVELOPMENT_VERSION.*$/s//    - DEVELOPMENT_VERSION=${version}/
+wq
+EOF
+}
+
+setPackagingNextReleaseVersion() {
+  local version="${1}"
+  local branchType="${2}"
+  local projectType="${3}"
+
+  if [[ "${projectType}" == "Packaging" ]]
+  then
+    if [[ "${branchType}" == "HotFix" ]]
+    then
+      setNextReleaseVersion "${version}.1"
+      setNextDevelopmentVersion "${version}.2-SNAPSHOT"
+    else
+      setNextReleaseVersion "${version}.A1-SNAPSHOT"
+      setNextDevelopmentVersion "${version}-SNAPSHOT"
+    fi
+  fi
 }
 
 incrementSchema() {
@@ -251,7 +297,12 @@ modifyPomVersion() {
 
   if [[ "${projectType}" == "Packaging" ]]
   then
-    pomVersion="${version}-SNAPSHOT"
+    if [[ "${branchType}" == "HotFix" ]]
+    then
+      pomVersion="${version}.1-SNAPSHOT"
+    else
+      pomVersion="${version}-SNAPSHOT"
+    fi
   else
     local versionMajor=`getVersionMajor "${version}"`
     if [[ "${versionMajor}" -lt 23 ]]
@@ -262,7 +313,7 @@ modifyPomVersion() {
         pomMinor=`getPomMinor`
         pomMinor=`increment "${pomMinor}"`
         pomVersion="${pomMajor}.${pomMinor}-SNAPSHOT"
-    else
+      else
         if [[ "${branchType}" == "Master" ]]
         then
           # Allow for two service packs before the next release off master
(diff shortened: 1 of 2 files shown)


scripts/dev/newReleaseBranch.sh
```diff
@@ -0,0 +1,680 @@
...
+     Creates the HotFix branch for 23.1.0, creates the 23.1.N ServicePack branch and modifies the master branch for
+     use by the next minor version 23.2.0.
+
+  2. After the release of 23.1.1 from the ServicePack branch release/23.1.N:
+     $ $0 23.1.1
+     Creates the HotFix branch for 23.1.1 and modifies the 23.1.N ServicePack branch for use by the next version 23.1.2
+
+  3. After the release of 23.1.2 from the ServicePack branch release/23.1.N
+     $ $0 23.1.2
+     Creates the HotFix branch for 23.1.2 and modifies the 23.1.N ServicePack branch for use by the next version 23.1.3
```
tpage-alfresco, line 36: I think this example is basically the same as the previous.

scripts/dev/newReleaseBranch.sh
```diff
@@ -0,0 +1,680 @@
...
+ROOT_DIR="${SCRIPT_DIR}/../../.."
+
+usage() {
+    cat << ???  1>&2;
+
+Creates HotFix and ServicePack (.N) branches for repo, share and packaging projects, and updates the master
+or ServicePack branches ready for the next release.
+
+Usage: $0 <hotFixVersion> [-v <masterVersion>] [-h] [-l]
+  <hotFixVersion>:  HotFix Branch to be created or modified.
```
tpage-alfresco, line 15: This is really the SP version. It's used for the HF branch name, but it's important that the parameter only has three parts.  e.g. Passing in 23.1.0.1 will break the script (or just cause a failure maybe - I didn't check).

scripts/dev/newReleaseBranch.sh
```diff
@@ -0,0 +1,680 @@
...
+wq
+EOF
+}
+
+setStartWithRealVersion() {
+  local version="${1}"
+
+  echo "${prefix}    set # ... start with real version ${version}"
+  ed -s .travis.yml &>${loggingOut} << EOF
+/.*start with real version.*$/s//    # Release version has to start with real version (${version}-....) for the docker image to build successfully./
```
tpage-alfresco, line 297: This whole method is just to update a comment correctly :)

tpage-alfresco (approved):
Looks really good to me. Nice work Alan!

scripts/dev/newReleaseBranch.sh
alandavis: Yes it is. The examples are linked. We generally do a major or minor release followed by TWO service pack releases and the next major or minor release. The for examples show this:
- 23.1.0
- 23.1.1
- 23.1.2   <-- the second SP that you correctly identified as basically the same as the previous one.
- 25.2.0   <-- under the year based versions 23 jumps to 25

scripts/dev/newReleaseBranch.sh
alandavis: I think you are correct to identify <hotFixVersion> in the Usage message could be improved. Changing it to:

Usage: $0 <releasedVersion> [-v <masterVersion>] [-h] [-l]
  <releasedVersion>:  The version just released. Used in the new HotFix Branch name release/<releasedVersion>.
                                    Must contain 3 integers separated by dots.

Internally the value is still assigned to the hotFixVersion variable as that still make sense.

scripts/dev/newReleaseBranch.sh
```diff
@@ -0,0 +1,681 @@
...
+  done < $pomFile
+  exit 1
+}
+
+getPomVersion() {
+  # Same as slower/simpler: "mvn help:evaluate -Dexpression=project.version"
+  readTopLevelTag version pom.xml
+}
+
+getPomMajor() {
```
mstrankowski, line 143: I was thinking about passing the version once as a parameter to these 3 methods, but it's probably negligible in performance while allowing cleaner usage overall

mstrankowski, line 143: I was thinking about passing the version once as a parameter to these 3 methods, but it's probably negligible in performance while allowing cleaner usage overall

alandavis: Given the whole script takes 1 minute, I would rather have more readable function names and cleaner code than have it run a few ms faster.

Optimisations are a good source of bugs normally, so I generally now only optimise existing code if I really have to. Yes it make sense to optimise the overall flow, as was done here. Originally the code was jumping from branch to branch doing the same change. It now does one branch at a time avoiding lots of git commands. Years ago everything had to be optimised as cpus and discs were slow. Not the case any more, but some teachers still get their students to optimise far too early. Having done both, I would rather have readable code.

scripts/dev/newReleaseBranch.sh
```diff
@@ -0,0 +1,681 @@
...
+  fi
+}
+
+incrementSchema() {
+  local schemaMultiple="${1}"
+
+  if [[ `getCurrentProject` == "alfresco-community-repo" && "${schemaMultiple}" -gt 0 ]]
+  then
+    local schema=`getSchema`
+    schema=`echo "(${schema} / ${schemaMultiple} + 1) * ${schemaMultiple}" | bc`
```
mstrankowski, line 358: What's happening here? If schemaMultiple is for example 100 (ServicePack) and current 
```
# Schema number
version.schema=17000
```
then it goes down to (17000/101) * 100 = ~16831
Are the multipliers in right order? I don't have much of knowledge about the release process itself or versioning, so it might be something obvious. If it is, then possibly a comment would be nice, as I have no idea what the schema multiplication is for (aside from changing version of schema to avoid clashes, but why such an operation is needed I don't have the slightest clue).

scripts/dev/newReleaseBranch.sh
```diff
@@ -0,0 +1,681 @@
...
+      if [[ "${branchType}" == "HotFix" ]]
+      then
+        pomMinor=`getPomMinor`
+        pomMinor=`increment "${pomMinor}"`
+        pomVersion="${pomMajor}.${pomMinor}"
+      else
+        if [[ "${branchType}" == "Master" ]]
+        then
+          # Allow for two service packs before the next release off master
+          pomMajor=`increment "${pomMajor}"`
```
mstrankowski, line 415: I understand the release of master ups the major version, but incrementing major version twice? Is a service pack release based on a major version? I would normally expect it to be based on minor version releases.

alandavis: Added a logger comment.

PR #2221, opened 2022-05-12
ACS-2985: Elasticsearch CMIS - a few basic +ve/-ve tests

montgolfiere:
- within acs-packaging (via V1 REST Search API using "cmis" query lang)

- note: currently Search API returns generic 500 (rather than 400) for invalid queries (whether syntactic or semantic)

### DIFF
--- a/tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
+++ b/tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
@@ -177,7 +177,7 @@ public class ElasticsearchCMISTests extends AbstractTestNGSpringContextTests
     @Test (groups = TestGroup.SEARCH)
     public void negative_basicCMISQuery_missingFrom()
     {
-        // note: ideally 400 but currently 500 :-(
+        // note: ideally 400 but currently 500 (also for Solr) :-(
 
         SearchRequest query1 = req("cmis", "SELECT *");
         searchQueryService.expectErrorFromQuery(query1, user1, HttpStatus.INTERNAL_SERVER_ERROR, "expecting FROM");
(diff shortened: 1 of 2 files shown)


krdabrowski (approved):

tpage-alfresco:
Do you know if the 500 errors are the same as for Solr?

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -172,6 +174,41 @@ public void checkInSyntax()
...
+        SearchRequest query2 = req("cmis", "SELECT * FROM");
+        searchQueryService.expectErrorFromQuery(query2, user1, HttpStatus.INTERNAL_SERVER_ERROR, "no viable alternative at input");
+    }
+
+    @Test (groups = TestGroup.SEARCH)
+    public void negative_basicCMISQuery_invalidType()
+    {
+        // note: ideally 400 but currently 500 :-(
+        SearchRequest query = req("SELECT * FROM cmis:unknown");
+        searchQueryService.expectErrorFromQuery(query, user1, HttpStatus.INTERNAL_SERVER_ERROR, "Unknown property");
```
tpage-alfresco, line 194: This error is pretty confusing. Does it mention `cmis:unknown` in the full error message?

tpage-alfresco (approved):
LGTM

montgolfiere:
> Do you know if the 500 errors are the same as for Solr? (As in - does Solr also return 500 for invalid queries, or does it return 400 instead?)

Yes, I did check previously and Solr also returns 500. We could log as minor tech debt to cleanup (POST /search API) in future Repo release.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
montgolfiere: Yes, this one is odd. It is also different to equivalent test at lower-level which gives a more useful "[Type is unsupported in query](https://github.com/Alfresco/alfresco-enterprise-repo/pull/809/files#diff-2466e672e08f8f55c9f5273d5f2a5555f148ce057aaf137da65bbd7df917f35cR95)" (note: private link)

TBC, but we may find this changes (gets fixed) as part of ACS-2920.

montgolfiere: I've updated the error messages to match currently what is returned. 

We may decide to revisit and improve, eg. if different from AFTS query lang &/or using CMIS query lang with Solr.

PR #2226, opened 2022-05-19
ACS-2805: Elasticsearch CMIS - test ORDER BY syntax

gevagel:
This PR needs [https://github.com/Alfresco/alfresco-enterprise-repo/pull/840](url) to be merged first in order for the tests to pass.

### DIFF
--- a/tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
+++ b/tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
@@ -597,14 +597,14 @@ public class ElasticsearchCMISTests extends AbstractTestNGSpringContextTests
     private List<String> getCreationDates(UserModel user, SiteModel site, FileModel... files)
     {
         return List.of(files)
-                .stream()
-                .map(FileModel::getCmisLocation)
-                .map(cmisLocation -> dataContent.usingUser(user).usingSite(site).getCMISDocument(cmisLocation))
-                .map(cmisDocument -> cmisDocument.<GregorianCalendar>getProperty("cmis:creationDate"))
-                .map(creationDateProperty -> creationDateProperty.<GregorianCalendar>getValue())
-                .map(GregorianCalendar::toInstant)
-                .map(Instant::toString)
-                .collect(toList());
+            .stream()
+            .map(FileModel::getCmisLocation)
+            .map(cmisLocation -> dataContent.usingUser(user).usingSite(site).getCMISDocument(cmisLocation))
+            .map(cmisDocument -> cmisDocument.<GregorianCalendar>getProperty("cmis:creationDate"))
+            .map(creationDateProperty -> creationDateProperty.<GregorianCalendar>getValue())
+            .map(GregorianCalendar::toInstant)
+            .map(Instant::toString)
+            .collect(toList());
     }
 
     private List<String> orderNames(String order, String... filename){
@@ -612,9 +612,14 @@ public class ElasticsearchCMISTests extends AbstractTestNGSpringContextTests
         if(order == DESC){
             orderedNames = orderedNames.stream().sorted(Comparator.reverseOrder()).collect(Collectors.toList());
         }
-        System.out.println("filename = " + filename);
-        System.out.println("orderedNames = " + orderedNames);
         return orderedNames;
     }
 
+    private FileModel uploadDocument(String filePath, UserModel user, SiteModel site) throws IOException
+    {
+        ClassPathResource toUpload = new ClassPathResource(filePath);
+        return dataContent.usingUser(user)
+            .usingSite(site)
+            .uploadDocument(toUpload.getFile());
+    }
 }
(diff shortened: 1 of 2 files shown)


tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/SearchQueryService.java
```diff
@@ -44,6 +45,12 @@ public void expectSomeResultsFromQuery(SearchRequest searchRequest, UserModel te
         expectResultsFromQuery(searchRequest, testUser, assertNotEmpty);
     }
 
+    public void expectResultsInOrder(SearchRequest searchRequest, UserModel user, List<String> expected)
```
tpage-alfresco, line 48: I think it will be easier to call the method if we use `String... expected` as in the other tests below.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -253,4 +266,13 @@ private FileModel createContent(String filename, String content, SiteModel site,
         return dataContent.usingUser(user).usingSite(site)
                           .createContent(fileModel);
     }
+
+    private List<String> orderNames(String... filename){
+        List<String> orderedNames = Arrays.stream(filename)
+                .sorted()
+                .collect(Collectors.toList());
```
tpage-alfresco, line 273: We can probably inline this if we're not going to reuse it.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -165,6 +169,15 @@ public void matchDocumentName()
         searchQueryService.expectResultsFromQuery(query, user1, FILE_0_NAME);
     }
 
+    @TestRail (description = "Check ORDER BY syntax works as expected", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
+    @Test (groups = TestGroup.SEARCH)
+    public void checkOrderByAscSyntax()
+    {
+        List<String> expectedList = orderNames(FILE_0_NAME, FILE_1_NAME, FILE_2_NAME);
```
tpage-alfresco, line 176: Could we also have a test for using `ORDER BY cmis:name DESC`?

montgolfiere: @gevagel Mentioned in slack

>  Tests for reverse sort are already in place.

Thanks. Can you add comment here (with link to code / commit) to confirm we test sort in both directions.

I'm sure Tom or you can then resolve this conversation.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -35,6 +40,8 @@ public class ElasticsearchCMISTests extends AbstractTestNGSpringContextTests
 {
     private static final String PREFIX = getAlphabeticUUID();
     private static final String SUFFIX = getAlphabeticUUID();
+    private static final String ASC = "ascending";
+    private static final String DESC = "descending";
```
tpage-alfresco, line 44: I think you're using these as an enum. Even simpler would be a boolean though.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -165,6 +169,15 @@ public void matchDocumentName()
         searchQueryService.expectResultsFromQuery(query, user1, FILE_0_NAME);
     }
 
+    @TestRail (description = "Check ORDER BY syntax works as expected", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
+    @Test (groups = TestGroup.SEARCH)
+    public void checkOrderByAscSyntax()
+    {
+        List<String> expectedList = orderNames(FILE_0_NAME, FILE_1_NAME, FILE_2_NAME);
+        SearchRequest query = req("cmis", "SELECT * FROM cmis:document WHERE cmis:name IN('" + FILE_0_NAME + "','" + FILE_1_NAME + "','" + FILE_2_NAME + "') ORDER BY cmis:name ASC");
```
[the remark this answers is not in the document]
gevagel: I think it would make sense to merge this piece of work first and then expand by adding tests for ordering of multiple fields. This build should now be green after [https://github.com/Alfresco/alfresco-enterprise-repo/pull/840](url) has been merged

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/SearchQueryService.java
```diff
@@ -121,6 +129,24 @@ private void assertNames(SearchResponse actual, String... expected)
         assertEquals(result, expectedList, "Unexpected search results.");
     }
 
+    private List<String> orderNames(boolean isAscending, String... filename){
+        List<String> orderedNames = Arrays.stream(filename).sorted().collect(Collectors.toList());
+        if(!isAscending){
+            orderedNames = orderedNames.stream().sorted(Comparator.reverseOrder()).collect(Collectors.toList());
+        }
+        return orderedNames;
+    }
```
montgolfiere, line 138: Do we need to sort here ? Why not just pass an expected list (in the expected order) ?

gevagel: Correct me if I'm wrong but I don't think we can do that since the file names have randomly generated prefixes. Thus, we need a method to do the sorting automatically.

tpage-alfresco (commented):
Belated :+1: (We already discussed via Slack last week anyway :) )

PR #2227, opened 2022-05-19
ACS-2800 Elasticsearch CMIS E2E test for cmis:objectTypeId field

damianujma:


tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -123,6 +123,25 @@ public void objectIdQuery()
         searchQueryService.expectResultsFromQuery(query, user1, FILE_0_NAME);
     }
 
+    @TestRail (description = "Check folders can be selected using cmis:objectTypeId.", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
+    @Test (groups = TestGroup.SEARCH)
+    public void objectTypeIdQuery()
+    {
+        SearchRequest query1 = req("cmis", "SELECT * FROM cmis:folder WHERE cmis:objectTypeId = 'cmis:folder'");
+        searchQueryService.expectResultsFromQuery(query1, user1,
+                // We expect the document library, ...
```
KikoUK, line 132: Looks good. 

Do we really need the comments? Code is quite descriptive :)

damianujma: Comments removed.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
```diff
@@ -123,6 +123,17 @@ public void objectIdQuery()
...
     }
 
+    @TestRail (description = "Check folders can be selected using cmis:objectTypeId.", section = TestGroup.SEARCH, executionType = ExecutionType.REGRESSION)
+    @Test (groups = TestGroup.SEARCH)
+    public void objectTypeIdQuery()
+    {
+        SearchRequest query1 = req("cmis", "SELECT * FROM cmis:folder WHERE cmis:objectTypeId = 'cmis:folder'");
+        searchQueryService.expectResultsFromQuery(query1, user1, "documentLibrary", FOLDER_0_NAME, FOLDER_1_NAME, user1.getUsername());
+
+        SearchRequest query2 = req("cmis", "SELECT * FROM cmis:folder WHERE cmis:objectTypeId = 'F:st:site'");
```
tpage-alfresco, line 133: Why does this work?

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchCMISTests.java
tpage-alfresco: Aha! I think the "and the user home for the user performing the query" is quite useful, because it's not obvious why a username is a folder. Anyway - I think this is present somewhere else in the file, so no need to change it back.

tpage-alfresco (approved):
We could potentially add a negative test too?  I was wondering about what happens with just `st:site` rather than `F:st:site`.  Not too important though - LGTM!

aonych (approved):

damianujma:
> We could potentially add a negative test too? I was wondering about what happens with just `st:site` rather than `F:st:site`. Not too important though - LGTM!

I will create another PR for a negative test.

PR #2234, opened 2022-05-25
ACS-2974 Zero Downtime Search Service Upgrade Test

pzhyland:


SaraAspery (commented):
Looks like this class has never had a licence header, but I wonder if it should?

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/upgrade/AvailabilityProbe.java
```diff
@@ -0,0 +1,135 @@
+package org.alfresco.elasticsearch.upgrade;
```
kmagdziarz, line 1: This class also doesn't have a licence header

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/upgrade/Config.java
```diff
@@ -0,0 +1,34 @@
+package org.alfresco.elasticsearch.upgrade;
```
kmagdziarz, line 1: Why creating second similar Interface to already created inner one? 
https://github.com/Alfresco/acs-packaging/pull/2234/files#:~:text=public%20interface-,ImagesConfig,-%7B

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/initialReindex/ElasticsearchInitialReindexingTests.java
```diff
@@ -1,6 +1,6 @@
 package org.alfresco.elasticsearch.initialReindex;
```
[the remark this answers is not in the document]
pzhyland: There are no copyright headers in the whole project. As we agreed, we won't add them right now but we will bring it to the management attention if we need them in all java files.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/upgrade/Config.java
pzhyland: It was added to decouple the upgrade tests components AcsEnv, Elasticsearch, ... from other parts of the system. In addition it makes it easier to run on a different version of docker images without changing some maven properties or env files.

kmagdziarz (approved):

SaraAspery (approved):
LGTM!

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/AlfrescoStackInitializer.java
```diff
@@ -297,4 +282,115 @@ protected GenericContainer createAlfrescoContainer()
...
+        @Override
+        public String getPostgreSQLImage()
+        {
+            return "postgres:" + envProperties.apply("POSTGRES_TAG");
+        }
+
+        @Override
+        public String getRepositoryImage()
+        {
+            return "alfresco/alfresco-content-repository:latest";
```
mstrankowski, line 377: All others seem to get a specific version except this one, it uses "latest", why?

mstrankowski: Nevermind, I started reviewing from the wrong angle, now I see it's a refactor, 👍

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/AlfrescoStackInitializer.java
```diff
@@ -297,4 +282,115 @@ protected GenericContainer createAlfrescoContainer()
                         "/usr/local/tomcat/webapps/alfresco/WEB-INF/classes/alfresco/search/elasticsearch/config/exactTermSearch.properties",
                         BindMode.READ_ONLY);
     }
+
```
mstrankowski, line 285: I kinda wonder if at this point it wouldn't be better to extract those into another class and provide them through configuration? Up to you, but it seems like a waste to separate it so neatly, just to not make it configurable (if different configs would ever be used, perhaps this is as static as they come)

pzhyland: I would leave it as it is. It's not needed for/related to the refactoring.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/upgrade/AvailabilityProbe.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private void start()
+    {
+        if (thread.isAlive()) return;
+        thread.start();
+    }
+
+    private void probing()
+    {
+        while (!stopRequested.get())
```
mstrankowski, line 57: I don't think I have enough context, but is there a possibility of locking up an additional thread running in endless loop if an exception occurs before "stop()" method is called?

pzhyland: Yes it is. This is why stop method should be called ;)

pzhyland: I changed it to a daemon thread so it won't block the jvm shutdown when it's not stopped correctly.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/upgrade/Elasticsearch.java
```diff
@@ -0,0 +1,153 @@
...
+    {
+        for (int i = 0; i < 3; i++)
+        {
+            waitFor("Elasticsearch Startup", ofMinutes(1), () -> {
+                try
+                {
+                    return !isIndexCreated();
+                } catch (IOException e)
+                {
+                    return false;
```
mstrankowski, line 128: While with "FileNotFoundException" in the inner method I can understand ignoring the exception and returning false, here I don't understand this approach. I get that after 1 minute it won't matter since a RuntimeException will be thrown from waitFor, but shouldn't exceptions other than FileNotFound actually immediately throw such a RuntimeException instead of locking up for 1 minute?

pzhyland: Nice catch! The outer loop is a leftover from a previous implementation I had in this method. I've removed it. Thank you.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/upgrade/FromSolrUpgradeTest.java
```diff
@@ -0,0 +1,192 @@
...
+    private static final String FILE_UPLOADED_AFTER_STARTING_LIVE_INDEXING = "after-live-indexing.pdf";
+    private static final String FILE_UPLOADED_BEFORE_SWITCHING_TO_ELASTICSEARCH = "before-switch.pdf";
+    private static final String FILE_UPLOADED_AFTER_SWITCHING_TO_ELASTICSEARCH = "after-switch.pdf";
+
+    @Test
+    public void testZeroDowntimeUpgradeFromSolrToElasticsearch() throws IOException
+    {
+        try (final UpgradeScenario scenario = new UpgradeScenario(getUpgradeScenarioConfig()))
+        {
+            final ACSEnv initialEnv = scenario.startInitialEnvWithSolrBasedSearchService();
```
mstrankowski, line 30: I'd try to split this test up with methods a bit, like a separate method "initialEnv = startInitialEnvironment()", "startMirroredEnvironment()", some methods like "verifyInitialEnvironmentIndexing()". It's a bit hard to read now when it's all on one level and we kind of lose the possibility of "documentation" through method names making it easier to read and separate code.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/AlfrescoStackInitializer.java
mstrankowski: After looking through code I feel like I know it a bit better, I understand that it kinda needs to be static anyway and passing it through configuration doesn't make much sense, but I still think it's separate enough to warrant being put in another file, even if only reason would be making the code less cluttered.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/upgrade/ACSEnv.java
```diff
@@ -0,0 +1,291 @@
...
+    private RepoHttpClient repoHttpClient;
+
+    private String metadataDumpToRestore;
+    private Path alfDataHostPath;
+    private boolean readOnlyContentStore;
+
+    public ACSEnv(Config cfg, final Network network, final String indexSubsystemName)
+    {
+        this.cfg = cfg;
+        postgres = newContainer(PostgreSQLContainer.class, cfg.getPostgreSQLImage())
```
mstrankowski, line 45: I personally would do something like "startEnvironmentContainers()" and within it "startPostgre", "startActiveMq", "startSharedFileStore" etc. It is understandable now though, so your choice.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/upgrade/FromSolrUpgradeTest.java
pzhyland: To be honest I don't know how to split it without overengineering it or mixing high level code with low level one.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/upgrade/ACSEnv.java
pzhyland: I moved containers creation to separate methods.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/upgrade/AvailabilityProbe.java
pzhyland: I introduced a change to stop the availability probe thread along with the environment it is running on.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/upgrade/FromSolrUpgradeTest.java
mstrankowski: They can git blame as we've talked to get the description of this case

PR #2249, opened 2022-07-06
Bump versions on several components

vrmoreira:


jottley:
@vrmoreira Do we still need this PR?

vrmoreira:
> @vrmoreira Do we still need this PR?

7.0.N is the only branch that wasn't updated. But, then again, it's not supported. IMHO, I would merge this PR in order to make every branch updated

jottley:
I would say that if we aren't going to be supporting it any more and we are far enough down the road from its EOL that we don't need the PR any more.

PR #2280, opened 2022-08-07
ACS-3425: Extreme test instability, to be stabilized in this issue

mstrankowski:


SaraAspery (approved):
If the test were to be disabled for longer then it might be better to use @Ignore, rather than comment out, just to make it easier to find ignored tests. Not sure it matters much for this one though!

mstrankowski:
I'll keep that in mind for future issues

PR #2302, opened 2022-09-27
ACS-3589 Use new artifact id for tas cmis.

tpage-alfresco:


dsibilio (approved):

tpage-alfresco:
Excluding failures which are on master the branch build passed.

PR #2310, opened 2022-10-07
Feature/acs 3644 - Add mechanism to support multiple databases in ACS TAS tests.

kmagdziarz:


tests/testcontainers-env/src/main/java/org/alfresco/tas/AlfrescoStackInitializer.java
```diff
@@ -248,6 +276,19 @@ private PostgreSQLContainer createPosgresContainer()
                                              .withStartupTimeout(Duration.ofMinutes(2));
     }
 
+
+    private MySQLContainer createMySqlContainer()
+    {
+        return (MySQLContainer) new MySQLContainer(getImagesConfig().getMySQLImage())
+                                            .withPassword("alfresco")
```
mstrankowski, line 283: might be worth to move to a static variable, seems that it occurs 3 times in this file

tests/testcontainers-env/src/main/java/org/alfresco/tas/AlfrescoStackInitializer.java
```diff
@@ -110,6 +111,33 @@ public void initialize(ConfigurableApplicationContext configurableApplicationCon
 
     }
 
+    private JdbcDatabaseContainer createDatabaseContainer()
+    {
+        switch (getImagesConfig().getDatabaseType())
+        {
+            case POSTGRESQL_DB:
+                setupDatabaseOptions("org.postgresql.Driver", "jdbc:postgresql://postgres:5432/alfresco");
```
mstrankowski, line 119: The first string could be added to enum values, don't know about the second. To be considered.

mstrankowski: If both are set in an enum, you could take it out of the switch, call setupDatabaseOptions(DatabaseType) and simply use retrieved values

tests/testcontainers-env/src/main/resources/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM quay.io/alfresco/alfresco-content-repository:latest
+
+ADD https://repo1.maven.org/maven2/mysql/mysql-connector-java/8.0.30/mysql-connector-java-8.0.30.jar /usr/local/tomcat/lib/mysql-connector-java-8.0.30.jar
```
mstrankowski, line 3: A bit of description in comment response please. I see that it takes the latest repository image (generated by maven during tests I suppose) and adds on the jar driver for mysql database, that's it here, right?

tests/testcontainers-env/src/main/java/org/alfresco/tas/AlfrescoStackInitializer.java
```diff
@@ -437,10 +478,27 @@ public String getPostgreSQLImage()
...
 
+        @Override
+        public String getMySQLImage()
+        {
+            return "mysql:" + envProperties.apply("MYSQL_TAG");
+        }
+
+        @Override
+        public DatabaseType getDatabaseType() {
+            String databaseTypeProperty = mavenProperties.apply("database.type");
```
mstrankowski, line 489: property name should be in static variable too imo

tests/testcontainers-env/src/main/java/org/alfresco/tas/AlfrescoStackInitializer.java
```diff
@@ -437,10 +478,27 @@ public String getPostgreSQLImage()
...
+
+            }
+            return DatabaseType.from(databaseTypeProperty);
+        }
+
         @Override
         public String getRepositoryImage()
         {
-            return "alfresco/alfresco-content-repository:latest";
+            return "alfresco-repository-databases";
```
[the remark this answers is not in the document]
kmagdziarz: https://github.com/Alfresco/acs-packaging/pull/2310#discussion_r1004487997
It is created in before-script stage in travis. I explained why we need this in linked comment.

tests/testcontainers-env/src/main/java/org/alfresco/tas/AlfrescoStackInitializer.java
kmagdziarz: It already exist for postgre

tests/testcontainers-env/src/main/java/org/alfresco/tas/AlfrescoStackInitializer.java
kmagdziarz: I wanted to stay consistent with the class.

mstrankowski (approved):

scripts/travis/build.sh
```diff
@@ -98,6 +98,11 @@ fi
 # Build the current project
 mvn -B -ntp -V -q install -DskipTests -Dmaven.javadoc.skip=true -Pbuild-docker-images -Pags ${REPO_IMAGE} ${SHARE_IMAGE}
 
+#Build alfresco image with jdbc drivers
+MYSQL_TAG=$(mvn help:evaluate -Dexpression=dependency.mysql.version -q -DforceStdout)
```
pzhyland, line 102: Does it make sense to add other drivers in this PR?

kmagdziarz: It may. But I assumed that better is to add one database in all places in one PR to prevent missing some parts.

tests/environment/docker-compose-mysql.yml
```diff
@@ -0,0 +1,12 @@
+services:
+  postgres:
```
pzhyland, line 2: shouldn't it be named a `database`

pzhyland (approved):
Just small comment n the mysql service name. Everything else looks good.

Don't forget to remove the SNAPSHOT dependencies from the main `pom.xml` file.

PR #2332, opened 2022-11-15
ACS-3963 - Add test with elasticsearch connector test agains MySQL database. 

kmagdziarz:


tests/environment/.env
```diff
@@ -20,7 +20,7 @@ SHARE_TAG=7.2.0
 DIGITAL_WORKSPACE_TAG=2.7.0
 #By default the connector tag is set to the value of `dependency.elasticsearch-shared.version` maven property.
 #Following env variable allow you to override this value. It is useful for testing. It should be left blank on the main branch.
-ES_CONNECTOR_TAG=
+ES_CONNECTOR_TAG=feature_ACS-3634-additional-jdbc-driver-loading-2034
```
mstrankowski, line 23: "It is useful for testing. It should be left blank on the main branch." in a comment right above it, is it a forgotten change? The value also seems kind of weird vs the purpose described in a comment.

tests/testcontainers-env/src/main/java/org/alfresco/tas/AlfrescoStackInitializer.java
```diff
@@ -154,6 +149,19 @@ public static void reindexEverything()
         }
     }
 
+    public static Map<String, String> getReindexEnvBasic()
```
mstrankowski, line 152: I really like taking out these common values into another method

tests/environment/.env
kmagdziarz: yup, it is for testing purpose, it has to be there until release of final ES connector as well as this PR should be merge once connector is released.

.travis.yml
```diff
@@ -234,12 +234,12 @@ jobs:
...
-#      vm:
-#        size: large
-#      script:
-#        - travis_wait 30 mvn -B install -ntp -pl ":content-repository-elasticsearch-test" -am -Pall-tas-tests,elastic -Denvironment=default -DrunBugs=false "-Dsearch.engine.type=elasticsearch" "-Ddatabase.type=mysql"
+    - name: "Elasticsearch TAS tests | MySQL (Search API)"
+      if: commit_message !~ /\[skip search\]/
+      vm:
+        size: large
+      script:
+        - travis_wait 30 mvn -B install -ntp -pl ":content-repository-elasticsearch-test" -am -Pall-tas-tests,elastic -Denvironment=default -DrunBugs=false "-Dsearch.engine.type=elasticsearch" "-Ddatabase.type=mysql"
```
mstrankowski, line 242: "-Ddatabase.type=mysql" 

Is the variable name correct?
I don't know if that was supposed to be the one forcing DB implementation type for elasticsearch-connector, but that might be different 
https://github.com/Alfresco/alfresco-elasticsearch-connector/pull/430/files
alfresco-elasticsearch-reindexing/src/main/resources/application.properties

mstrankowski: If that's not it, then just resolve the comment

.travis.yml
```diff
@@ -253,12 +253,10 @@ jobs:
...
-#    - name: "Elasticsearch Basic Auth TAS tests | MySQL"
-#      if: commit_message !~ /\[skip search\]/
-#      before_script:
-#        - docker build -t alfresco-repository-databases:latest tests/environment
-#      script:
-#        - travis_wait 30 mvn -B install -ntp -pl ":content-repository-elasticsearch-test" -am -Pall-tas-tests,elastic-basic-auth -Denvironment=default -DrunBugs=false "-Dsearch.engine.type=elasticsearch" "-Ddatabase.type=mysql"
+    - name: "Elasticsearch Basic Auth TAS tests | MySQL"
+      if: commit_message !~ /\[skip search\]/
+      script:
+        - travis_wait 30 mvn -B install -ntp -pl ":content-repository-elasticsearch-test" -am -Pall-tas-tests,elastic-basic-auth -Denvironment=default -DrunBugs=false "-Dsearch.engine.type=elasticsearch" "-Ddatabase.type=mysql"
```
mstrankowski, line 259: same as above

pzhyland (approved):

PR #2341, opened 2022-11-24
Feature/acs 4046 maria db support

mstrankowski:


tests/testcontainers-env/pom.xml
```diff
@@ -45,10 +45,20 @@
...
+            <artifactId>mariadb</artifactId>
+            <version>${dependency.testcontainers.version}</version>
+        </dependency>
         <dependency>
             <groupId>org.testcontainers</groupId>
             <artifactId>elasticsearch</artifactId>
             <version>${dependency.testcontainers.version}</version>
         </dependency>
+
+        <dependency>
```
mstrankowski, line 59: please check the location, couldn't set the scope of dependency as test, but since it's located in tests only should be fine

pzhyland: I think we should keep additional drivers in a `test` scope. What do you think of moving these drivers to the `tests/tas-elasticsearch/pom.xml`.

pzhyland (approved):

kcichonczyk (approved):

PR #2343, opened 2022-11-28
Feature/acs 4042 mssql tests support

kcichonczyk:


kmagdziarz (approved):

tests/environment/docker-compose-mssql.yml
```diff
@@ -0,0 +1,12 @@
+version: "3"
+
+services:
+  database:
+    image: mcr.microsoft.com/mssql/server:2019-CU9-ubuntu-16.04
```
pzhyland, line 5: Shouldn't we use `MSSQL_TAG` here?

pzhyland (approved):

tests/environment/docker-compose-mssql.yml
```diff
@@ -0,0 +1,12 @@
...
+services:
+  database:
+    image: mcr.microsoft.com/mssql/server:${MSSQL_TAG}
+    mem_limit: 1512m
+    environment:
+      - SA_PASSWORD=Alfresco1
+      - ACCEPT_EULA=Y
+    command: /opt/mssql/bin/sqlservr
+    ports:
+      - "1433:1433"
```
mstrankowski, line 12: new line

tests/testcontainers-env/src/main/java/org/alfresco/tas/AlfrescoStackInitializer.java
```diff
@@ -298,6 +303,16 @@ private MariaDBContainer createMariaDBContainer()
 
     }
 
+    private JdbcDatabaseContainer createMsSqlContainer()
+    {
+        return new MSSQLServerContainer<>(getImagesConfig().getMsSqlImage())
+                .acceptLicense()
```
mstrankowski, line 309: If you decide to move usage of some values to enum (as above) then you could use them here, like DatabaseType.MSSQL.getPassword()

tests/testcontainers-env/src/main/java/org/alfresco/tas/DatabaseType.java
```diff
@@ -34,6 +53,28 @@ public String getUrl()
         return url;
     }
 
+    public String getSpringDatasourceUrl()
+    {
+        if(this.equals(DatabaseType.MSSQL_DB))
+        {
+            return getUrl().replace("\\", "");
```
mstrankowski, line 60: you first use these values, perhaps to escape a character, then you replace them with nothing? Please explain

kcichonczyk: Url passed to container definition as JAVA_OPTS needs to have ';' escaped, therefore '\\' there. Url passed as SPRING_DATASOURCE_URL doesn't need that and is incorrect when left with '\\'.

tests/environment/docker-compose-mssql.yml
kcichonczyk: added.

tests/testcontainers-env/src/main/java/org/alfresco/tas/DatabaseType.java
kcichonczyk: changed according to suggestions

PR #2380, opened 2023-01-30
ACS-4527 Fix upstream repo linking

dsibilio:
Fix linking to other upstream projects when their version is set to `-SNAPSHOT`.

The issue is related to a cached `GITHUB_TOKEN` within the local git config, due to the way the `checkout` action works. It's possible to turn this behaviour off via the `persist-credentials: false` input.

Furthermore, we don't seem to need this token caching functionality anyways since we make sure to run the `configure-git-author` action with the appropriate parameters whenever we need a valid git identity to perform certain operations.

_Example of a **failed run**: https://github.com/Alfresco/acs-packaging/actions/runs/4044397436/jobs/6954489020#step:5:165
Example of a **successful run** (after the fix): https://github.com/Alfresco/acs-packaging/actions/runs/4044437652/jobs/6955842727#step:5:164_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,8 @@ jobs:
       !contains(github.event.head_commit.message, '[skip tests]')
     steps:
       - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
       - uses: Alfresco/alfresco-build-tools/.github/actions/get-build-info@v1.32.0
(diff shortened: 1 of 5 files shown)


tpage-alfresco (approved):
Nice find!

damianujma (approved):

aonych (approved):

PR #2382, opened 2023-01-31
ACS-4155 CRUD for categories with ES.

tpage-alfresco:


krdabrowski (approved):

tests/testcontainers-env/src/main/java/org/alfresco/tas/AlfrescoStackInitializer.java
```diff
@@ -102,7 +103,7 @@ public void initialize(ConfigurableApplicationContext configurableApplicationCon
         startOrFail(transformCore, transformRouter);
 
         // We don't want Kibana to run on our CI, but it can be useful when investigating issues locally.
-        if (Objects.equals(System.getProperty("kibana"), "true"))
+        if (getSystemProperty("kibana", "false").equals("true"))
```
mpichura, line 106: Flipping equals here would be safer in case of NPE.
```suggestion
        if ("true".equals(getSystemProperty("kibana", "false")))
```

mpichura (approved):
LGTM but at some point you need to remove SNAPSHOTS in POMs.

tests/testcontainers-env/src/main/java/org/alfresco/tas/AlfrescoStackInitializer.java
tpage-alfresco: `getSystemProperty` returns the default value if the property isn't specified rather than `null` - i.e. it will be `false`.

pom.xml
```diff
@@ -10,12 +10,12 @@
         <groupId>org.alfresco</groupId>
         <artifactId>alfresco-enterprise-repo</artifactId>
         <relativePath>../alfresco-enterprise-repo/pom.xml</relativePath>
-        <version>20.58</version>
+        <version>20.59-SNAPSHOT</version>
```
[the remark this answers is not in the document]
tpage-alfresco: This is now resolved as the enterprise repo change has been merged :+1:

PR #2390, opened 2023-02-14
ACS-4021 Add E2E tests for path updates.

tpage-alfresco:


tpage-alfresco:
There are several tests that seem to fail still. Currently I have an additional reindex step included which fixes one of the tests.

https://github.com/Alfresco/acs-packaging/pull/2390/files#diff-602b535bed137f9ed931f64150bfa3aeda1104ab4b810d02f79b8a4024b77b93R274

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/reindexing/PathUpdateTests.java
```diff
@@ -0,0 +1,396 @@
...
+
+    /**
+     * Create a path to a file or folder via the category hierarchy.
+     *
+     * @param nodeNames The ordered list of node names from the root category to the node that was categorised.
+     * @return An absolute path through the categories to the specified node.
+     */
+    private static String categoryPath(String... nodeNames)
+    {
+        return "/cm:categoryRoot/cm:general_classifiable/cm:" + Arrays.stream(nodeNames).collect(joining("/cm:"));
```
tpage-alfresco, line 394: This should be `cm:generalclassifiable`

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/reindexing/PathUpdateTests.java
```diff
@@ -0,0 +1,396 @@
...
+                RestCategoryModel parentCategory = createCategory(restClient, dataUser.getAdminUser(), "-root-");
+                RestCategoryModel childCategory = createCategory(restClient, dataUser.getAdminUser(), parentCategory.getId());
+                FileModel fileModel = FileModel.getRandomFileModel(TEXT_PLAIN);
+                FileModel testFile = dataContent.usingUser(testUser).usingSite(testSite).createContent(fileModel);
+                restClient.authenticateUser(testUser).withCoreAPI().usingNode(testFile)
+                          .linkToCategory(RestCategoryLinkBodyModel.builder().categoryId(childCategory.getId()).create());
+
+                Step.STEP("Update the parent category name and check the file's paths are updated.");
+                String newCategoryName = parentCategory.getName() + "_updated";
+                restClient.authenticateUser(testUser).withCoreAPI()
```
mpichura, line 225: This should be done with admin user.

mpichura (commented):
Had a few remarks.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/reindexing/PathUpdateTests.java
```diff
@@ -0,0 +1,462 @@
...
+    protected SearchQueryService searchQueryService;
+
+    private UserModel testUser;
+    private SiteModel testSite;
+
+    /** Map from test method name to the set of named properties used by the test. */
+    private Map<String, Map<String, Object>> propertyMaps;
+
+    /** All the "given", "when" and "then" steps for each test, grouped according to the test name. */
+    private Map<String, Triple<Supplier<Map<String, Object>>, Function<Map<String, Object>, Map<String, Object>>, Consumer<Map<String, Object>>>> testSteps = Map.of(
```
krdabrowski, line 90: Maybe it's worth considering hiding this crazy definition in some class, like `TestSteps extends Map<...>`. This could allow also to hide the logic of tests methods verification, particular steps execution, etc.

krdabrowski (approved):
Overall LGTM - nice job!

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/reindexing/PathUpdateTests.java
tpage-alfresco: Yeah - that sounds good. Potentially we could have any number of step initialisation steps separated by waits, followed by a verification step for each test.

tests/tas-elasticsearch/src/test/resources/test-suites/elasticsearch-parallel-suite.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd">
+<suite name="Parallel REST API tests for Elasticsearch" verbose="1" parallel="true" thread-count="100">
```
krdabrowski, line 3: Why thread count is set to 100, it's an upper limit?

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/parallel/RetryAnalyzer.java
```diff
@@ -0,0 +1,40 @@
...
+import org.testng.ITestResult;
+
+/**
+ * Sometimes REST tests fail when run in parallel with:
+ * <pre>
+ *     java.lang.IllegalStateException: Invalid use of BasicClientConnManager: connection still allocated.
+ *     Make sure to release the connection before allocating another one.
+ * </pre>
+ * Other times restassured fails with <code>java.util.ConcurrentModificationException</code>.
+ * If we detect a test failed due to these timing issues then we rerun it up to three times.
```
krdabrowski, line 17: Those exceptions are thrown due to shared user and site?

krdabrowski:
Nice solution and usage of TestNG parallel testing capabilities. I like also yours retry analyzer - nice discovery!

tests/tas-elasticsearch/src/test/resources/test-suites/elasticsearch-parallel-suite.xml
tpage-alfresco: Oh - this was an accident. I meant to leave it at the default (5), but I was playing around with it locally.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/parallel/RetryAnalyzer.java
tpage-alfresco: These exceptions seem to be due to multiple tests using RestAssured in parallel. I ran the tests about 30 times locally (with various settings) and these are the only two exceptions I saw which were not application issues.

krdabrowski: RestAssured probably is not entirely thread-safe, so I think you're right. Thanks for explaining.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/parallel/PathUpdateTests.java
```diff
@@ -0,0 +1,333 @@
...
+     */
+    private void waitForOneSecond()
+    {
+        try
+        {
+            TimeUnit.SECONDS.sleep(1);
+        }
+        catch (InterruptedException e)
+        {
+            LOGGER.warn("Failed to wait for full second.");
```
mpichura, line 324: I'm not sure if this is a general rule but at least in multi-threaded apps (don't know how the parallelism works here) there was a rule of thumb not to ignore `InterruptedException` and IIRC, as the simplest mean just call `Thread.currentThread().interrupt();`.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/parallel/RetryAnalyzer.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class RetryAnalyzer implements IRetryAnalyzer
+{
+    private static Logger LOGGER = LoggerFactory.getLogger(RetryAnalyzer.class);
+    private static int RETRY_LIMIT = 3;
+    private int retryNumber = 0;
```
mpichura, line 23: I am not sure if `retry` method here can be called from multiple threads but if so, it would be much safer to use `AtomicInteger` instead of `int` as `retryNumber`

mpichura (approved):
LGTM - hopefully this will help to make our tests more reliable. 
I had two remarks regarding possible multi-threadeding issues.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/parallel/PathUpdateTests.java
tpage-alfresco: I was able to remove these sleep requests completely as we have fixed the underlying issue in the ESC.

PR #2393, opened 2023-02-20
ACS-4139 Produce Multi-Arch Docker images

wojtekswieton:


dsibilio (approved):

scripts/ci/build.sh
```diff
@@ -6,6 +6,19 @@ pushd "$(dirname "${BASH_SOURCE[0]}")/../../"
...
+   case $option in
+      m) # Docker image build mode
+         IS_MULTI_ARCH=$OPTARG;;
+     \?) # Invalid option
+         echo "Error: Invalid option"
+         exit;;
+   esac
+done
+
+BUILD_PROFILE=$([[ "${IS_MULTI_ARCH}" = "multiarch" ]] && echo "build-multiarch-docker-images" || echo "build-docker-images")
```
dsibilio, line 20: I think we could refactor this a little bit, more or less like this:

```suggestion
usage() {
    echo "Builds the upstream projects first, then the current one." 1>&2;
    echo 1>&2;
    echo "Usage: $0 [-m]" 1>&2;
    echo "  -m: Flag to build Docker images with multi-architecture" 1>&2;
    echo "  -h: Display the usage information" 1>&2;
    exit 1;
}

while getopts "mh" option; do
   case $option in
      m)
        DOCKER_BUILD_PROFILE=build-multiarch-docker-images
        ;;
      h)
        usage
        ;;
   esac
done

BUILD_PROFILE=${DOCKER_BUILD_PROFILE:-build-docker-images}
```

dsibilio: **NOTE:** this causes the invocations of `build.sh -m multiarch` to be simplified to `build.sh -m`

.github/workflows/master_release.yml
```diff
@@ -37,18 +37,27 @@ jobs:
       !(failure() || cancelled()) &&
       !contains(github.event.head_commit.message, '[skip docker_latest]') &&
       github.ref_name == 'master' && github.event_name != 'pull_request'
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
```
dsibilio, line 44: Can you remind me what's the use of this?

tests/pipeline-all-amps/pom.xml
```diff
@@ -16,17 +16,27 @@
         <module>share</module>
     </modules>
 
+    <properties>
+        <skipDockerImage>true</skipDockerImage>
```
dsibilio, line 20: I'm curious why did we need to set this to true, when the sub-modules are set to false?

dsibilio (approved):
LGTM, great work ✅ Left a few questions though.

tests/pipeline-all-amps/pom.xml
damianujma: We need this property to prevent an attempt to build an image for `pipeline-all-amps` module.

.github/workflows/master_release.yml
damianujma: We need local registry as a docker container to build, pull and push multi-arch base images there. https://docs.docker.com/build/ci/github-actions/local-registry/

damianujma (approved):

PR #2397, opened 2023-02-24
Remove unnecessary files from hosted runners to free up some disk space

pzhyland:


dsibilio (approved):
Since we're at it, it would be nice to sync the alfresco-build-tools version used in https://github.com/Alfresco/acs-packaging/blob/master/.github/workflows/master_release.yml to v1.35.2 as well ✔️

goleksyalfresco (approved):

PR #2400, opened 2023-02-28
OPSEXP-1944 Switch to new acs license access & secret keys

slohe1:
OPSEXP-1944

PR #2401, opened 2023-02-28
Use S3 pipeline secrets only where they are really needed

gionn:
Noticed this while reviewing https://github.com/Alfresco/acs-packaging/pull/2400


dsibilio (approved):
LGTM, assuming tests pass.

PR #2413, opened 2023-03-17
OPSEXP-2046 Add base build number to avoid conflicts with previous travis build number

gionn:
Ref: OPSEXP-2046

This can be reverted once moving away from `7.4.0-XXXX` and `23.1.0-XXXX`

### DIFF
--- a/.github/workflows/master_release.yml
+++ b/.github/workflows/master_release.yml
@@ -20,6 +20,7 @@ env:
   QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
   QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
   GITHUB_ACTIONS_DEPLOY_TIMEOUT: 60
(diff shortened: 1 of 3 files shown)


.github/workflows/master_release.yml
```diff
@@ -20,6 +20,7 @@ env:
   QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
   QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
   GITHUB_ACTIONS_DEPLOY_TIMEOUT: 60
+  BASE_BUILD_NUMBER: 10000
```
gionn, line 23: Screenshot 2023-03-17 at 10 16 54

alxgomz (approved):

dsibilio (approved):
image

I think this may still be needed if the next version is `23.1.x`, until `23.1.x` where `x` > `0`... we could revert it then.

PR #2415, opened 2023-03-17
ACS-4139 Temporarily disable multi-arch for pipeline-all-amps images

damianujma:


dsibilio (approved):
LGTM, can we please add a comment saying that we can restore this once we migrate to Flux2 in terraform-alfresco-pipeline due to Flux1 not supporting the multi-arch images?

**NOTE:** feel free to test this directly on master, if it doesn't work as intended then let's revert it immediately and simply update terraform-alfresco-pipeline/develop manually from time to time.

PR #2430, opened 2023-04-04
 ACS-4462: Mutual TLS e2e test

kmagdziarz:


tests/tas-mtls/src/test/java/org/alfresco/rest/MtlsRestTest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.slf4j.Logger;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.testng.AbstractTestNGSpringContextTests;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.BeforeSuite;
+
+@ContextConfiguration ("classpath:alfresco-mtls-context.xml")
+public abstract class MtlsRestTest extends AbstractTestNGSpringContextTests
```
mstrankowski, line 21: Do I understand it correctly, that this test is only verifying if Repository is alive and working?

tests/tas-mtls/src/test/java/org/alfresco/rest/mtls/TransformServiceTest.java
```diff
@@ -0,0 +1,47 @@
...
+        FileModel fileModel = FileModel.getRandomFileModel(TEXT_PLAIN);
+        FileModel testFile = dataContent.usingUser(adminUser).usingResource(testFolder).createContent(fileModel);
+
+        restClient.authenticateUser(adminUser).withCoreAPI().usingNode(testFile).createNodeRendition("pdf");
+        restClient.assertStatusCodeIs(HttpStatus.ACCEPTED);
+
+        String status = restClient.withCoreAPI().usingNode(testFile).getNodeRenditionUntilIsCreated("pdf").getStatus();
+        Assert.assertEquals(status, "CREATED");
+
+        restClient.authenticateUser(adminUser).withCoreAPI().usingNode(testFile).createNodeRendition("doclib");
```
mstrankowski, line 41: Just keep in mind that in general such cases would be better split into 2 different tests.
Here it's fine since we don't exactly test the renditions, more the mTLS communication, so fail anywhere is good.

tests/tas-mtls/src/test/resources/default.properties
```diff
@@ -0,0 +1,2 @@
+# dataprep related
+alfresco.port=8082
```
mstrankowski, line 2: A question, I remember all the woes faced to fix that port issue, buut, couldn't we change the port in docker compose instead to fit to the one provided by default?

tests/tas-mtls/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+      </properties>
+   </profile>
+   </profiles>
+
+   <dependencies>
+      <dependency>
+         <groupId>com.google.code.gson</groupId>
+         <artifactId>gson</artifactId>
+         <scope>test</scope>
+      </dependency>
```
mstrankowski, line 35: Are we using Gson?

tests/environment/docker-compose-mtls.yml
```diff
@@ -0,0 +1,299 @@
...
+  digital-workspace:
+    image: quay.io/alfresco/alfresco-digital-workspace:${DIGITAL_WORKSPACE_TAG}
+    mem_limit: 128m
+    environment:
+      APP_CONFIG_AUTH_TYPE: "BASIC"
+      BASE_PATH: ./
+      APP_BASE_SHARE_URL: "http://localhost:8080/workspace/#/preview/s"
+
+  control-center:
+    image: quay.io/alfresco/alfresco-admin-app:7.6.1
```
mstrankowski, line 256: It's the only one with a non-variable version, why?

tests/environment/docker-compose-mtls.yml
```diff
@@ -0,0 +1,299 @@
...
+        -Ddb.password=alfresco
+        -Ddb.url=jdbc:postgresql://postgres:5432/alfresco
+        -Dsolr.host=solr6
+        -Dsolr.port=8983
+        -Dsolr.secureComms=secret
+        -Dsolr.sharedSecret=secret
+        -Dsolr.base.url=/solr
+        -Dindex.subsystem.name=solr6
+        -Dalfresco.restApi.basicAuthScheme=true
+        -Dtransform.service.enabled=false
```
mstrankowski, line 50: This property is defined twice, also, if possible, please group the properties so they aren't in a scattered order :) (sorry :D)

tests/tas-mtls/src/test/resources/default.properties
mstrankowski: I meant the exposed port in docker compose, in the "ports" section where we decide what it should be, not the property value. Though granted they would need to change too.

scripts/ci/generate_keystores.sh
```diff
@@ -0,0 +1,31 @@
+#! /bin/bash
+#! /bin/bash
+
+# SETTINGS
+# Alfresco Format: "classic" / "current" is supported only from 7.0
+ALFRESCO_FORMAT=current
+
+#Contains directory settings
+source ${GITHUB_WORKSPACE}/alfresco-ssl-generator/ssl-tool/utils.sh
```
pzhyland, line 9: Please avoid such dependency (`GITHUB_WORKSPACE`) on the execution environment. The better approach would be to set some variable (with a meaningful name) before executing the script.

tests/environment/docker-compose-mtls.yml
```diff
@@ -0,0 +1,295 @@
...
+      CLIENT_SSL_TRUST_STORE_PASSWORD: "password"
+      CLIENT_SSL_TRUST_STORE_TYPE: "JCEKS"
+    ports:
+      - "8099:8099"
+    volumes:
+      - shared-file-store-volume:/tmp/Alfresco/sfs
+      - ${GITHUB_WORKSPACE}/keystores/sharedFileStore/sharedFileStore.truststore:/sharedFileStore.truststore
+      - ${GITHUB_WORKSPACE}/keystores/sharedFileStore/sharedFileStore.keystore:/sharedFileStore.keystore
+
+  share:
```
pzhyland, line 189: Do we need it for the test?

tests/environment/docker-compose-mtls.yml
```diff
@@ -0,0 +1,295 @@
...
+  activemq:
+    image: quay.io/alfresco/alfresco-activemq:${ACTIVEMQ_TAG}
+    mem_limit: 1g
+    ports:
+      - 8161:8161 # Web Console
+      - 5672:5672 # AMQP
+      - 61616:61616 # OpenWire
+      - 61613:61613 # STOMP
+
+  digital-workspace:
```
pzhyland, line 243: Do we need it for the test?

tests/environment/docker-compose-mtls.yml
```diff
@@ -0,0 +1,295 @@
...
+    image: quay.io/alfresco/alfresco-digital-workspace:${DIGITAL_WORKSPACE_TAG}
+    mem_limit: 128m
+    environment:
+      APP_CONFIG_AUTH_TYPE: "BASIC"
+      BASE_PATH: ./
+      APP_BASE_SHARE_URL: "http://localhost:8080/workspace/#/preview/s"
+
+  control-center:
```
pzhyland, line 251: Do we need it for the test?

scripts/ci/generate_keystores.sh
kmagdziarz: I'm going to unify that script in https://alfresco.atlassian.net/browse/ACS-5005
Now it scattered across 5 repositories.

tests/environment/docker-compose-mtls.yml
```diff
@@ -0,0 +1,295 @@
...
+      - control-center
+    ports:
+      - "8080:8080"
+    links:
+      - digital-workspace
+      - alfresco
+      - share
+      - control-center
+
+  sync-service:
```
pzhyland, line 274: Do we need it for the test?

tests/environment/docker-compose-mtls.yml
```diff
@@ -0,0 +1,295 @@
...
+
+  control-center:
+    image: quay.io/alfresco/alfresco-admin-app:7.6.1
+    mem_limit: 128m
+    environment:
+      APP_CONFIG_PROVIDER: "ECM"
+      APP_CONFIG_AUTH_TYPE: "BASIC"
+      BASE_PATH: ./
+
+  proxy:
```
pzhyland, line 259: Do we need it for the test?

tests/environment/docker-compose-mtls.yml
```diff
@@ -0,0 +1,295 @@
...
+      CLIENT_SSL_KEY_STORE_TYPE: "JCEKS"
+
+      CLIENT_SSL_TRUST_STORE: "file:/sharedFileStore.truststore"
+      CLIENT_SSL_TRUST_STORE_PASSWORD: "password"
+      CLIENT_SSL_TRUST_STORE_TYPE: "JCEKS"
+    ports:
+      - "8099:8099"
+    volumes:
+      - shared-file-store-volume:/tmp/Alfresco/sfs
+      - ${GITHUB_WORKSPACE}/keystores/sharedFileStore/sharedFileStore.truststore:/sharedFileStore.truststore
```
pzhyland, line 186: Don't depend on the execution environment.

scripts/ci/generate_keystores.sh
pzhyland: It's not about unifying the key generations but it's rather about the dependency on the runtime environment.

tests/tas-mtls/src/test/java/org/alfresco/rest/mtls/TransformServiceTest.java
```diff
@@ -0,0 +1,47 @@
...
+    private UserModel adminUser;
+
+    @BeforeClass (alwaysRun = true)
+    public void dataPreparation()
+    {
+        adminUser = dataUser.getAdminUser();
+    }
+
+    @Test
+    public void testRenditionWithMTLSEnabledTest()
```
pzhyland, line 27: I am wondering if there is a way to verify if the mTLS is really enabled. Right now we use the Rest API and even if there is no mTLS is not enabled (or is broken) the test will pass. Maybe `MtlsRestTest` could just check if T-Router is enforcing mTLS?

scripts/ci/generate_keystores.sh
kmagdziarz: Okey, I will also deal with that runtime dependency in others repository.

tests/environment/docker-compose-mtls.yml
```diff
@@ -0,0 +1,295 @@
...
+        -Ddb.password=alfresco
+        -Ddb.url=jdbc:postgresql://postgres:5432/alfresco
+        -Dsolr.host=solr6
+        -Dsolr.port=8983
+        -Dsolr.secureComms=secret
+        -Dsolr.sharedSecret=secret
+        -Dsolr.base.url=/solr
+        -Dindex.subsystem.name=solr6
+        -Dalfresco.restApi.basicAuthScheme=true
+        -Dtransform.service.enabled=false
```
pzhyland, line 50: We should enable transform service (and probably disable local transformers) to involve all ATS components.

pzhyland (changes requested):
There are still some changes needed. Especially around involving all ATS components.

tests/environment/docker-compose-mtls.yml
kmagdziarz: container removed

tests/tas-mtls/src/test/java/org/alfresco/rest/mtls/TransformServiceTest.java
kmagdziarz: Added

tests/environment/docker-compose-mtls.yml
kmagdziarz: Done.

tests/tas-mtls/src/test/java/org/alfresco/rest/MtlsRestTest.java
```diff
@@ -0,0 +1,103 @@
...
+    public void showEndTestInfo(Method method)
+    {
+        LOGGER.info(String.format("*** ENDING Test: [%s] ***", method.getName()));
+    }
+
+    @Test
+    public void checkIfMtlsIsEnabledForTransformService() throws IOException
+    {
+        CloseableHttpResponse transformServiceResponse;
+        transformServiceResponse = client.execute(new HttpGet("http://localhost:8090"));
```
pzhyland, line 78: Try to get the schema (I think it should be `https`) the hostname and the port from the configuration (it should be the same source as for the `assertServerIsOnline`)

pzhyland: My bad. It's not the repo url but the t-router, but still `https` should be used.

mstrankowski (approved):

PR #2437, opened 2023-04-11
GHA Workflow on master/release - Fix publish stage 

evasques:
Newly created tag was not being able to be checked-out


dsibilio:
Most likely https://github.com/Alfresco/acs-community-packaging/blob/master/.github/workflows/ci.yml#L207 will need the same kind of fix.

evasques:
@dsibilio I'll add it there also

PR #2449, opened 2023-04-21
ACS-5026 Add JIRA integration to the ARM64 health check - to close

damianujma:
Adding JIRA integration to the ARM64 health check workflow to create issues within ACS in case the build fails and close issues in case the build passes.


.github/workflows/arm64.yml
```diff
@@ -83,6 +86,14 @@ jobs:
...
             --ignore=tests/scenarios/test_transformation.py \
             -k "not test_search"
+      - run: echo "TEST_FAILURE=true" >> $GITHUB_ENV
+        if: failure()
+      - name: Create JIRA issue
+        if: always()
+        run: |
+          pip install -r ./scripts/ci/requirements.txt
+          ./scripts/ci/create_jira_issue.py
+        shell: bash
```
dsibilio, line 96: Is this really needed?

.github/workflows/arm64.yml
```diff
@@ -83,6 +86,14 @@ jobs:
             --ignore=tests/scenarios/test_sync.py \
             --ignore=tests/scenarios/test_transformation.py \
             -k "not test_search"
+      - run: echo "TEST_FAILURE=true" >> $GITHUB_ENV
+        if: failure()
+      - name: Create JIRA issue
+        if: always()
+        run: |
+          pip install -r ./scripts/ci/requirements.txt
+          ./scripts/ci/create_jira_issue.py
```
dsibilio, line 95: Did we test this by running the ARM64 ACS Packaging CI via `gh workflow run .github/workflows/arm64.yml  --ref this/branch-name`? I believe this shouldn't work given that we're checking out an older tag for tests which won't contain these files (hence why the workarounds related to `dtas-config.json` and the pipeline-all-amps docker-compose can be seen in previous steps.

In case it doesn't work (as I'd expect), we could temporarily apply the same workaround for these files as well, saving these files before checking out the test tag and moving them in their appropriate location after the test tag has been checked out. Once a new -Ax tag of ACS is available we can remove the workaround as they will be there since now on.

scripts/ci/create_jira_issue.py
```diff
@@ -0,0 +1,206 @@
...
+        if key:
+            logging.info(f"Closing issue {key}")
+            jira.issue_add_comment(key, get_comment())
+            jira.set_issue_status(key, "Done", fields={'fixVersions': [{'name': 'none'}]})
+
+    else:
+        logging.info("Test failed, create issue")
+        # Checks before trying creating jira issue
+        # 1. Is test declared in python dict
+        # 2. Is build-user have access to project
```
dsibilio, line 202:
```suggestion
        # Checks before trying to create jira issue
        # 1. Is test declared in python dict
        # 2. Does build-user have access to project
```

.github/workflows/arm64.yml
```diff
@@ -11,13 +11,16 @@ env:
   MAVEN_USERNAME: ${{ secrets.NEXUS_USERNAME }}
   TAS_ENVIRONMENT: ./tests/environment
   TAS_SCRIPTS: ../alfresco-community-repo/packaging/tests/scripts
+  JIRA_TOKEN: ${{ secrets.JIRA_TOKEN }}
```
dsibilio, line 14: This could be an env var of the specific step that needs it.

.github/workflows/arm64.yml
```diff
@@ -11,13 +11,16 @@ env:
...
   AWS_REGION: eu-west-1
   DTAS_VERSION: v1.1.3
 
 jobs:
   arm64_health_check:
     name: ARM64 Health Check
     runs-on: [self-hosted, linux, arm64]
+    env:
+      BUILD_WEB_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
```
dsibilio, line 23: Could this be an env var of the specific step that needs it, or does it need to be declared at the job-level to work?

scripts/ci/requirements.txt
```diff
@@ -0,0 +1 @@
+atlassian-python-api==3.18.0
```
dsibilio, line 1: It would be good to make a subfolder within `scripts/ci` e.g.: `scripts/ci/jira` and move both the python script and this file inside it, otherwise it can be hard to understand what this `requirements.txt` is referring to if we happen to add more python scripts to this directory.

dsibilio (changes requested):
Good work. Left a few comments.

_P.S.: the commit history is very dirty, it doesn't make this PR much harder to review as there are no previous CI executions linked to commits I'd be interested in, nor the changes are particularly extensive. And as long as we remember to `Squash and Merge`, then all is good. BUT, as a general note, in cases where the commit history is so dirty it would be best to either rebase the current branch to master or simply make a new clean branch from master and apply your changes there if it's easier for you. This would make it so that the commit history actually makes sense._

.github/workflows/arm64.yml
dsibilio: Note: alternatively we could checkout the `master` branch again before running the JIRA integration, via the `checkout` action.

gionn (commented):
would be very valuable to have it at least wrapped in a custom action, if not having it in a dedicated repo where we can add later also some tests given that the logic here is not really trivial

would also love to refactor the onein the terraform pipeline and have a single create jira to handle all the features

damianujma:
I'm closing this PR due to the fact I created new PR https://github.com/Alfresco/acs-packaging/pull/2450 with a new clean branch from master.

.github/workflows/arm64.yml
damianujma: No. I removed it.

Ref. https://github.com/Alfresco/acs-packaging/pull/2450

.github/workflows/arm64.yml
damianujma: I used the `checkout` action.
Ref. https://github.com/Alfresco/acs-packaging/pull/2450

scripts/ci/create_jira_issue.py
damianujma: Done.
Ref. https://github.com/Alfresco/acs-packaging/pull/2450

.github/workflows/arm64.yml
damianujma: Done.
Ref. https://github.com/Alfresco/acs-packaging/pull/2450

.github/workflows/arm64.yml
damianujma: I believe this can be env var of the specific step. Done.
Ref. https://github.com/Alfresco/acs-packaging/pull/2450

scripts/ci/requirements.txt
damianujma: Done.
Ref. https://github.com/Alfresco/acs-packaging/pull/2450

PR #2454, opened 2023-04-27
ACS-5025 Remove now unnecessary ARM64 workflow workarounds

dsibilio:
Removing the ARM64 workflow workarounds which are now unnecessary due to the fact that the latest available ACS tag is now containing the expected files.

**Test run**: https://github.com/Alfresco/acs-packaging/actions/runs/4819040407

### DIFF
--- a/.github/workflows/arm64.yml
+++ b/.github/workflows/arm64.yml
@@ -49,14 +49,7 @@ jobs:
           echo "TEST_GIT_TAG=$TEST_GIT_TAG" >> $GITHUB_ENV
       - name: "Prepare test config"
         run: |
-          # NOTE: copy files from master branch as temporary workaround until a tag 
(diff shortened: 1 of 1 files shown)


.github/workflows/arm64.yml
```diff
@@ -119,4 +112,3 @@ jobs:
           JIRA_TOKEN: ${{ secrets.JIRA_TOKEN }}
           JOB_NAME: arm64_health_check
           TEST_FAILURE: ${{ needs.arm64_health_check.outputs.test_failure }}
-
```
damianujma, line 122: Can we remain this empty line?

damianujma: I check it locally and it also looks fine on my end 👍

damianujma (approved):

mikolajbrzezinski (approved):

PR #2489, opened 2023-06-07
ACS-4063 Remove provenance layer from Docker images

dsibilio:
Testing the removal of the provenance attestation layer which causes issues on quay.io, such as the expire label not working and unknown manifests being shown.

**Note**: will be tested on master, reverted if there are issues/doesn't work.


gionn (approved):

dsibilio:
Seems to work ✅

PR #2491, opened 2023-06-08
ACS-4800 ES tests for ANCESTOR.

tpage-alfresco:


tests/testcontainers-env/src/main/java/org/alfresco/tas/MavenPropertyHelper.java
```diff
@@ -35,7 +37,8 @@ public static String getMavenProperty(String key)
     public static Properties loadMavenProperties()
     {
         Properties properties = new Properties();
-        try (FileReader reader = new FileReader(MAVEN_PROPERTIES_FILE))
+        InputStream inputStream = MavenPropertyHelper.class.getClassLoader().getResourceAsStream(MAVEN_PROPERTIES_FILE);
```
mpichura, line 40: I think it would be better if you included `inputStream` in try with resources.
```suggestion
        try (
        InputStream inputStream = MavenPropertyHelper.class.getClassLoader().getResourceAsStream(MAVEN_PROPERTIES_FILE);
        Reader reader = new InputStreamReader(inputStream)
        )
```

tpage-alfresco: I don't think it makes any difference, since the actual file pointer will be closed anyway. Do you think it's clearer like this though?

mpichura: Having looked at `InputStreamReader::close` implementation I see that it closes the `inputStream` so it doesn't make any difference but I wouldn't have to dive into it if it was included in the clause, so maybe it is clearer then. 
Anyway - feel free to take it or leave it.

mpichura (approved):
LGTM

azakrzewski-hy (approved):

PR #2494, opened 2023-06-16
ACS-4800 E2E tests for parent and primary parent.

tpage-alfresco:


tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/reindexing/ElasticsearchPathIndexingTests.java
```diff
@@ -169,6 +169,20 @@ public void testAncestorWithWorkspaceReference() {
...
+        searchQueryService.expectResultsFromQuery(query, testUser, testFolders.get(2).getName());
+    }
+
+    @Test (groups = TestGroup.SEARCH)
+    public void testParent()
+    {
+        SearchRequest query = req("PARENT:\"" + testFolders.get(1).getNodeRef() + "\" AND name:*");
+        searchQueryService.expectResultsFromQuery(query, testUser, testFolders.get(2).getName());
+    }
+
```
azakrzewski-hy, line 185: Do I understand it right that calling for nested folder (1) we expect to get direct parent of it? Like (folder 0 (folder 1 (folder 2)))

tpage-alfresco: folder0 contains folder1 which contains folder2. The query says "Give me everything that has the parent "folder1", and the result is just folder2.

The `name:*` simply forces the query to go to the index, as otherwise the query might get handled by the DB instead.

azakrzewski-hy (approved):
LGTM

PR #2509, opened 2023-07-05
Add share verification and alfresco verification with redirect ignore

mstrankowski:


mstrankowski:
The only tests that failed are completely unrelated to the actual test case modified in a simple manner. I'll treat those failures as intermittent failures.

PR #2520, opened 2023-07-21
Adding label do docker images of pipeline_... to allow for actual dep…

mstrankowski:
…recation and removal of such images when restricting time from command line


mstrankowski:
For push there are only intermittent failures, branch lived for like 20 minutes so PR build won't matter much, changes are unrelated to failures, they just enable setting the validity duration of pipeline docker images uploaded to quay. By default those never expire, now you can make them expire after a set amount of time (as is for alfresco-repo and alfresco-share)

PR #2527, opened 2023-07-31
ACS-5221 Add ARM64 search tests

mikolajbrzezinski:


mikolajbrzezinski:
Ive left the docker logs to be displayed even on job failure since when they exit after creation but don't cause build failure we should still be able to inspect them IMO. If you think its unnecessary let me know, I'll switch it back.

.github/workflows/arm64.yml
```diff
@@ -70,7 +70,6 @@ jobs:
           curl -v --user admin.pipeline@alfresco.com:admin http://localhost:8080/alfresco/api/discovery
       - name: "Dump all Docker containers logs"
         uses: Alfresco/alfresco-build-tools/.github/actions/docker-dump-containers-logs@v1.41.0
-        if: failure() && steps.setup-env.outcome == 'failure'
```
dsibilio, line 73: I don't think there's any significant advantage in having the logs always uploaded, especially if everything went well.

I definitely see room for improvement though, as things are right now if the tests fail we don't have the Docker logs to better look into the failure, which is a shame.

Please let's move the `Dump all docker container logs` step **after** the tests are run, and let's run it in case of failures where either the `setup-env` or `run-tests` step fail.

dsibilio (changes requested):

.github/workflows/arm64.yml
mikolajbrzezinski: Sure, that's better, ill change that around

dsibilio (approved):

aonych (approved):

PR #2539, opened 2023-08-10
ACS-5734 Fix Elasticsearch TAS tests failure

viepovsky:
After changing memory settings, there are no longer `OutOfMemoryError` logs, however this did not fix failing tests.
There are randomly failing tests with AssertionError message: `Maximum retry period reached, test failed.`

To minimise failing tests I changed the `RetryAnalyzer` class and its `retry` method to re-run the test if AssertionError happens with given message. This has reduced failing tests decently. 

Apart from above, there were often failing tests from `testSiteUseCasesForCreateModifyDeleteSite()` test with `Maximum retry period reached, test failed.` message . Even after re-run from `RetryAnalyzer` this was failing due to not dumping data from previous test run. So I basically added some lines to check if there are existing files, sites and delete them at the begining of this test. This solution is working and now these tests are passing even if they fall, they are re-runned and they pass.

All this decently reduced failing ES TAS tests but sometimes there are still tests failing with message: `Maximum retry period reached, test failed.`. It may happen because of: 
-the two additional re-runs did not help, so the test end up with the same error message or 
-the failing test is not covered with RetryAnalyzer so it is not re-running even if falls.


tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/reindexing/ElasticsearchSiteIndexingTests.java
```diff
@@ -103,14 +112,23 @@ public void testSiteUseCasesForCreateModifyDeleteSite()
         // Remove the automatically created Sample Site
         deleteSite(SAMPLE_SITE_ID);
 
+        //Sometimes this test may fail, so if previous data exists it must be deleted before the next run
```
dsibilio, line 115: How does this test's retry happen exactly?

As far as I can see the Elasticsearch TestNG suite that is triggering the `reindexing` tests does not use the `org.alfresco.elasticsearch.parallel.RetryAnnotationTransformer` that kicks in the `RetryAnalyzer`: https://github.com/Alfresco/acs-packaging/blob/e00c708955af7733decb72b2ccf3254bc5a4167f/tests/tas-elasticsearch/src/test/resources/test-suites/elasticsearch-suite.xml#L5.

viepovsky: I think that these tests use the `elastic` profile set up in `pom.xml` file:
https://github.com/Alfresco/acs-packaging/blob/b620a647bd03a44056675463bdf2a0b089beb105/tests/tas-elasticsearch/pom.xml#L19
And inside `standard-elasticsearch-suites.xml` there is `elasticsearch-parallel-suite.xml` which has `<listener class-name="org.alfresco.elasticsearch.parallel.RetryAnnotationTransformer" />`

It looks from the logs like this:
error

dsibilio: Seems to be a bug/unwanted behaviour of TestNG when merging test suites in one combined XML file: https://sqa.stackexchange.com/questions/37457/testng-listener-from-one-xml-file-is-applied-to-another.

So it was most likely unwanted/unknown by the person that introduced this listener + merged suite file. Either way, if it helps with making the tests a bit more stable I'm not against keeping it.

viepovsky: Yes it helps a lot in this case

dsibilio: cc @tpage-alfresco so he's aware about this behaviour _(in case he wasn't, I definitely wasn't)_ 👍

mikolajbrzezinski (approved):

aonych (approved):

damianujma (approved):

PR #2564, opened 2023-09-06
Test if Java 11 tests can detect issues with amp

pzhyland:


pom.xml
```diff
@@ -21,7 +21,7 @@
 
         <alfresco.s3connector.version>5.1.0</alfresco.s3connector.version>
         <alfresco.azure-connector.version>3.2.0</alfresco.azure-connector.version>
-        <alfresco.salesforce-connector.version>2.4.0</alfresco.salesforce-connector.version>
+        <alfresco.salesforce-connector.version>2.4.1</alfresco.salesforce-connector.version>
```
kavitshah-gl, line 24: The version 2.4.1 is not compatible with jdk11 so the test should fail? I was trying to look at the Github action pipeline run for the same but there were not failures. Am I looking into something else or understanding wrong?

pzhyland: Because of the branch name not all of the jobs were executed. Domenico triggered the workflow manually and it failed https://github.com/Alfresco/acs-packaging/actions/runs/6106795837/job/16572516061#step:8:652

kavitshah-gl (approved):
approved

PR #2566, opened 2023-09-07
ACS-5865 Use latest OJDBC driver

pzhyland:


dsibilio (approved):

kavitshah-gl (approved):
Seems fine, the test job failed in pipeline (specific to oracle) but I think that is intermediate failure and nothing to do with Oracle driver upgrade.

PR #2578, opened 2023-09-22
Update ya-pmd-scan to 2.0.4.

tpage-alfresco:


tpage-alfresco:
Ignore one failing job since this change did not affect it.

PR #2582, opened 2023-09-26
ACS-4953 ES highlighting tests.

tpage-alfresco:


azakrzewski-hy (approved):
LGTM

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchHighlightingTests.java
```diff
@@ -0,0 +1,166 @@
...
+    static final String CONTENT_B = """
+            The lazy dog sleeps under the quick brown fox. The middle of the document is quite long:
+            Lorem ipsum dolor sit amet. In veniam tempore hic provident sunt et distinctio velit et reprehenderit
+            officiis id sapiente omnis et quisquam aliquam et porro perferendis. In sequi placeat ut quaerat voluptatem
+            ea consequuntur impedit aut eaque enim aut atque rerum.
+            The end of the document mentions the dog again!
+            """;
+
+    @Autowired
+    ServerHealth serverHealth;
```
krdabrowski, line 56: Are those fields used anywhere else? Shouldn't they be marked as `private`?

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchHighlightingTests.java
```diff
@@ -0,0 +1,166 @@
...
+    DataSite dataSite;
+    @Autowired
+    SearchQueryService searchQueryService;
+
+    UserModel user;
+    SiteModel site;
+    FileModel fileA;
+    FileModel fileB;
+
+    @BeforeClass (alwaysRun = true)
```
krdabrowski, line 71: I'm wondering about the space after annotation name - I treat annotations like methods and I'm not placing a space before brackets in such a case. I can start doing it, but I think we all should agree on such a convention.

tpage-alfresco: I upgraded my IDE recently, so maybe it's only just started doing this. Looking through the existing ACS code then it seems to be overwhelmingly in favour of not having a space so I'll update my settings to remove this space.

tests/tas-elasticsearch/src/test/java/org/alfresco/elasticsearch/ElasticsearchHighlightingTests.java
tpage-alfresco: I saw somewhere the idea that it's slightly harmful adding `private` modifiers to test classes because we reduce the readability without adding any useful protection (test classes should be independent anyway).

I like the idea of having less boilerplate code, but tbh I'm not convinced that I "see" the private modifier on fields anyway.

What do you think?

krdabrowski: I have this kind of habit to add everywhere `private` modifier, so I won't forget abut it in the future. In the other hand you are probably right - IIRC JUnit 5 has dropped `private` modifier for test methods.
I'm fine with this approach - we could agree on that and treat it as a convention - @mpichura WDYT?

krdabrowski (approved):
Everything except `-SNAPSHOT` in pom LGTM ;)

tpage-alfresco:
The issue with the PMD scan should be fixed in this PR: https://github.com/Alfresco/ya-pmd-scan/pull/7

PR #2618, opened 2023-10-26
[ACS-5768] Switch to the Jakarta compatible Jolokia web application

kmagdziarz:


jottley:
@pzhyland Do we still need this PR?

pzhyland:
Let's leave it for now. We will try to introduce this change during the Maintenance Week.

PR #2630, opened 2023-11-09
Disabling jobs requiring new license

pzhyland:
These jobs require real 23.2 license. Since we are not going to generate new licenses for this version (23.1 license should be used for 23.X). These jobs should be restored after introducing new versioning scheme to the repo.


dsibilio (approved):
😔

tpage-alfresco (approved):
I'll add a note to ACS-6286 mentioning this.

PR #2631, opened 2023-11-10
ACS-6282 Switch to a disposable EC2 hosted runner for ARM64 tests

dsibilio:
Changing the approach we use to do ARM64 tests in ACS Packaging.

### **Before**
We rely on a permanently deployed EC2 runner which is easily subject to degradation over time and incurs in issues which usually result in having to reboot the runner manually or recreate it altogether. Also, this kind of runner doesn't allow parallel builds and must be cleaned thoroughly before/after the execution as it's not disposable.

### **After**
We rely on an on-demand EC2 runner that is created, set-up, used to run the tests and is finally destroyed - on the fly. This also allows parallel runs with multiple runners with no additional effort.

Also, I commented out the `test_adw_reachability` as it was failing, most likely due to the fact that ADW is emulated and therefore very slow => the test is also irrelevant, it should be enabled once again as native ARM64 ADW becomes available.

Also, I tuned some of the timeout parameters for renditions/requests as I noticed one instance of flaky rendition tests, after this change I didn't notice them anymore.

> ℹ️ **NOTE**
> _It is technically possible to shave off ~1 minute from the workflow execution by executing the "Install required software" section once and freezing it in the source AMI. At the moment I don't think it's worth it due to 1. ideally the real end-game solution for this would be https://github.com/github/roadmap/issues/836 2. it would hide all the details that are needed to make the runner work and goes against the spirit of runners being disposable machines not necessarily tailored to our purposes by default._

**Sample run:** https://github.com/Alfresco/acs-packaging/actions/runs/6828040152 ✅

### DIFF
--- a/.github/workflows/arm64.yml
+++ b/.github/workflows/arm64.yml
@@ -108,7 +108,8 @@ jobs:
           echo "TEST_GIT_TAG=$TEST_GIT_TAG" >> $GITHUB_ENV
       - name: "Prepare test config"
         run: |
-          git checkout $TEST_GIT_TAG
+          #Temporarily test with more generous timeouts due to flaky tests
+          #git checkout $TEST_GIT_TAG
           mvn clean install -B -ntp -DskipTests -f tests/pipeline-all-amps/repo/pom.xml
           cat tests/pipeline-all-amps/repo/target/dtas/dtas-config.json
       - name: "Set up the environment"
--- a/.github/workflows/arm64.yml
+++ b/.github/workflows/arm64.yml
@@ -46,7 +46,7 @@ jobs:
           pre-runner-script: |
             set -ex
             sudo apt-get update
-            sudo apt-get install ca-certificates curl gnupg
+            sudo apt-get install ca-certificates curl gnupg -y
             sudo install -m 0755 -d /etc/apt/keyrings
             curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
             sudo chmod a+r /etc/apt/keyrings/docker.gpg
@@ -55,7 +55,7 @@ jobs:
               "$(. /etc/os-release && echo "$VERSION_CODENAME")" stable" | \
               sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
             sudo apt-get update
-            sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin git unzip curl -y
+            sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin git unzip -y
 
   arm64_health_check:
     name: "ARM64 Health Check"
--- a/.github/workflows/arm64.yml
+++ b/.github/workflows/arm64.yml
@@ -43,19 +43,18 @@ jobs:
               {"Key": "Name", "Value": "arm64-acs-packaging-gh-runner"},
               {"Key": "GitHubRepository", "Value": "${{ github.repository }}"}
             ]
-          pre-runner-script: |
-            set -ex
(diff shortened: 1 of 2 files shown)


aonych (approved):
LGTM 👍 Good work!

.github/workflows/arm64.yml
```diff
@@ -15,19 +15,78 @@ env:
...
+            "deb [arch="$(dpkg --print-architecture)" signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu \
+            "$(. /etc/os-release && echo "$VERSION_CODENAME")" stable" | \
+            sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+          sudo apt-get update
+          # Additionally install git, zip and unzip (required by SDKMAN to install Maven), python3
+          sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin git zip unzip python3 python3-pip -y
+          # Allow usage of "docker-compose" rather than "docker compose" in bash scripts
+          ln -f -s /usr/libexec/docker/cli-plugins/docker-compose /usr/local/bin/docker-compose
+          # Workaround due to $HOME not being set causing issues with settings.xml installation in setup-build-java
+          mkdir -p /root/.m2
```
gionn, line 78: you are using sudo in the previous commands but if this is not failing here without sudo it means that current user is root so you don't need sudo in any of the previous commands

.github/workflows/arm64.yml
```diff
@@ -15,19 +15,78 @@ env:
...
+          aws-access-key-id: ${{ secrets.AWS_EC2_GITHUB_RUNNER_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_EC2_GITHUB_RUNNER_SECRET_ACCESS_KEY }}
+          aws-region: eu-west-2
+      - name: "Start EC2 runner"
+        id: start-ec2-runner
+        uses: machulav/ec2-github-runner@4e0303de215db88e1c489e07a15ca4d867f488ea
+        with:
+          mode: start
+          github-token: ${{ secrets.BOT_GITHUB_TOKEN }}
+          ec2-image-id: ami-0ec31c3e3b729bb6d
```
gionn, line 37: what is this AMI? add a comment to give an hint

.github/workflows/arm64.yml
```diff
@@ -15,19 +15,78 @@ env:
...
+          ec2-instance-type: c6g.2xlarge
+          subnet-id: subnet-0777c70a4cd9ce944
+          security-group-id: sg-0f89a325d9eb147cb
+          aws-resource-tags: >
+            [
+              {"Key": "Name", "Value": "arm64-acs-packaging-gh-runner"},
+              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"}
```
gionn, line 44: can you use some "standard" tags? https://alfresco.atlassian.net/wiki/spaces/TECH/pages/1266352366/AWS+tags+for+resource+identifications

gionn (changes requested):

.github/workflows/arm64.yml
dsibilio: yeah, the `sudo`s here are redundant as the `root` user is the only user that can be engaged by this action. I'll get rid of the unnecessary `sudo`s 👍

.github/workflows/arm64.yml
dsibilio: good idea, thanks for sharing the page as I wasn't aware of it. Will do 💯

mikolajbrzezinski (approved):

damianujma (approved):

.github/workflows/arm64.yml
dsibilio: tags

gionn (approved):

PR #2637, opened 2023-11-16
ACS-6304 Implement SAST Pipeline Scan

mikolajbrzezinski:
The Share scan issue has been forwarded to Veracode and will be tracked and implemented in a separate ticket when the problem is resolved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
       - name: "Clean Maven cache"
         run: bash ./scripts/ci/cleanup_cache.sh
 
-  veracode_sast:
+  veracode_sast_acs:
     name: "Pipeline SAST Scan"
     runs-on: ubuntu-latest
     if: >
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -60,6 +60,56 @@ jobs:
...
+          bash ./scripts/ci/init.sh
+          bash ./scripts/ci/build.sh
+      - name: "Run SAST Scan"
+        uses: veracode/Veracode-pipeline-scan-action@v1.0.10
+        with:
+          vid: '${{ secrets.VERACODE_API_ID }}'
+          vkey: '${{ secrets.VERACODE_API_KEY }}'
+          file: "distribution/target/alfresco.war"
+          fail_build: true
+          project_name: acs-packaging-acs
```
dsibilio, line 92: Why is the project name `acs-packaging-acs` instead of e.g.: `acs-packaging`?

.github/workflows/ci.yml
```diff
@@ -38,7 +38,7 @@ env:
   DTAS_VERSION: v1.2.2
 
 jobs:
-  veracode:
+  veracode_sca:
```
dsibilio, line 41: We're missing the condition change we planned to do across all our repos to guarantee that SCA scans would also run on PRs, check https://github.com/Alfresco/alfresco-enterprise-repo/pull/1442/files as a reference.

dsibilio (changes requested):

.github/workflows/ci.yml
mikolajbrzezinski: Because usually there is no need for distinction, but here there is both acs and share in one project so to distinguish them from one another I added the -acs part here, and for share we should add -share. If you have a better idea feel free to suggest it

dsibilio (approved):

aonych (approved):

mikolajbrzezinski:
The Share scan issue has been forwarded to Veracode and will be tracked and implemented in a separate ticket when the problem is resolved.

PR #2656, opened 2023-11-24
MNT-24063 Migrate to GHA

dsibilio:
As per title.


.github/workflows/master_release.yml
```diff
@@ -0,0 +1,253 @@
+name: Master/Release branch workflow
+
+on:
+  push:
+    branches:
+      - master
+      - release/**
+  schedule:
```
damianujma, line 8: I think we can remove this `schedule` trigger because it works only on `master` branch.

dsibilio: I'd rather keep it so that we keep the least amount of discrepancies between this release branch and the other already-migrated release branches which have the same unnecessary `schedule` in place. Making it easier for backports/cherry-picks with no harm since it does nothing on branches other than the default one.

aonych (approved):

PR #2667, opened 2024-01-09
ACS-6406 new release branch script adjusted to new ACS versioning schema

mpichura:
This python script mimics the behavior of the shell script we used for new branch creation after ACS release but it is adjusted to new ACS versioning schema. 
Additionally, script gives opportunity to create release branches before the release is done so that we don't need to freeze the code. 
It is written in python mostly because it gives more control of processing output and error handling.
As I have revisited python after many years, I am open to any suggestions.

### DIFF
--- a/scripts/dev/newReleaseBranch.py
+++ b/scripts/dev/newReleaseBranch.py
@@ -170,8 +170,8 @@ def update_ci_yaml(filename, project, rel_version, dev_version):
 
     if text:
         logger.debug("Setting RELEASE_VERSION, DEVELOPMENT_VERSION (%s, %s) in %s ci.yml" % (rel_version, dev_version, project))
-        update_line(text, release_version_match, rel_version)
-        update_line(text, development_version_match, dev_version)
+        update_line(text, release_version_match, rel_version + "-A1")
+        update_line(text, development_version_match, dev_version + "-A1-SNAPSHOT")
         with open(filename, 'w') as file:
             file.writelines(text)
 
--- a/scripts/dev/newReleaseBranch.py
+++ b/scripts/dev/newReleaseBranch.py
@@ -26,7 +26,6 @@
 # - set ACS version properties in main POM to X.Y.1
 # - set POM versions to  X.Y.1.1-SNAPSHOT
 # - set scm-tag in main POM to HEAD
-# - increment schema by 1 in repository.properties
 # - set version.revision to 1 in version.properties (test resources)
 # 6. acs-community-packaging
 # - not created as we do not release hot fixes for community version
@@ -34,7 +33,7 @@
 # Create ServicePack branches for the released version (for X.Y.Z release it will be release/X.Y+1 eg., create release/23.3 for 23.2.0 release)
 # 1. acs-packaging:
 # - set RELEASE_VERSION to X.Y+1.0-A1, DEVELOPMENT_VERSION to X.Y+1.0-A2-SNAPSHOT in master_release.yml
-# - set POM versions to X.Y+1.0-SNAPSHOT
+# - set POM versions to X.Y+1.0-A1-SNAPSHOT
 # - set scm-tag in main POM to HEAD
 # 2. enterprise-share
 # - set scm-tag in main POM to HEAD
@@ -56,7 +55,7 @@
 # - set version.revision to Z+1 in version.properties (test resources)
 # 6. acs-community-packaging
 # - set RELEASE_VERSION to X.Y+1.0-A1, DEVELOPMENT_VERSION to X.Y+1.0-A2-SNAPSHOT in ci.yml
-# - set POM versions to X.Y+1.0-SNAPSHOT
+# - set POM versions to X.Y+1.0-A1-SNAPSHOT
 # - set scm-tag in main POM to HEAD
 # - set comm-repo dependency in main POM to X.Y+1.0.1
 # - set comm-share dependency in main POM to X.Y+1.0.1
@@ -65,7 +64,7 @@
 # 1. acs-packaging:
 # - set RELEASE_VERSION to <next_development_version>-A1 passed as script argument or X.Y+1.0-A1 (if <next_development_version> not passed),
 #   DEVELOPMENT_VERSION to <next_development_version>-A2-SNAPSHOT or X.Y+1.0-A2-SNAPSHOT (if <next_development_version> not passed) in master_release.yml
-# - set POM versions to <next_development_version>-SNAPSHOT or X.Y+1.0-SNAPSHOT (if <next_development_version> not passed)
+# - set POM versions to <next_development_version>-A1-SNAPSHOT or X.Y+1.0-A1-SNAPSHOT (if <next_development_version> not passed)
 # - set scm-tag in main POM to HEAD
 # 2. enterprise-share
 # - set scm-tag in main POM to HEAD
@@ -88,6 +87,7 @@
 # 6. acs-community-packaging
 # - set RELEASE_VERSION to <next_development_version>-A1 passed as script argument or X.Y+1.0-A1 (if <next_development_version> not passed),
 #   DEVELOPMENT_VERSION to <next_development_version>-A2-SNAPSHOT or X.Y+1.0-A2-SNAPSHOT (if <next_development_version> not passed) in master_release.yml
+# - set POM versions to <next_development_version>-A1-SNAPSHOT or X.Y+1.0-A1-SNAPSHOT (if <next_development_version> not passed)
 # - set scm-tag in main POM to HEAD
 # - set comm-repo dependency in main POM to <next_development_version>.1 or X.Y+1.0.1 (if <next_development_version> not passed)
 # - set comm-share dependency in main POM to <next_development_version>.1 or X.Y+1.0.1 (if <next_development_version> not passed)
@@ -123,7 +123,7 @@ parser.add_argument('-n', '--next_dev', metavar='x.y.z', help='next development 
 parser.add_argument('-v', '--verbose', action='store_true', help='print out verbose processing information')
 parser.add_argument('-s', '--skip_push', action='store_true', help='skip git push')
 parser.add_argument('-t', '--test_branches', action='store_true', help='use test branches')
-parser.add_argument('-c', '--cleanup', action='store_true', help='cleanup local release branches (experimental)')
+parser.add_argument('-c', '--cleanup', action='store_true', help='cleanup local test release branches (experimental)')
 parser.add_argument('-a', '--ahead', action='store_true', help='create branches ahead of release (experimental)')
 parser.add_argument('-z', '--trace', action='store_true', help='trace processing information')
 
@@ -365,8 +365,8 @@ def calculate_increment(version, next_dev_ver):
     if is_version_bumped(version, next_dev_ver, 1):
         logger.debug("Increment by 100")
         return 100
-    logger.debug("Increment by 1")
-    return 1
+    logger.debug("Increment by 0")
+    return 0
 
 
 def update_ent_repo_acs_label(project, version, branch_type):
@@ -408,10 +408,11 @@ def get_cmd_exec_result(cmd_args):
         raise
 
 
-def set_versions(project, version, profiles: list[str]):
+def set_versions(project, version, branch_type):
+    profiles = ["dev"] if "packaging" in project else ["ags"]
     switch_dir(project)
     if "packaging" in project:
-        snapshot_ver = version + "-SNAPSHOT"
+        snapshot_ver = version + "-SNAPSHOT" if branch_type == HOTFIX else version + "-A1-SNAPSHOT"
     else:
         ver = version.split(".")
         if len(ver) == 4:
@@ -499,8 +500,7 @@ def calculate_version(project):
 
 
 def update_project(project, version, branch_type):
-    profiles = ["dev"] if "packaging" in project else ["ags"]
-    set_versions(project, version, profiles)
+    set_versions(project, version, branch_type)
     update_scm_tag('HEAD', project)
     next_dev_ver = get_next_dev_version(branch_type)
     if project == ACS_PACKAGING:
--- a/scripts/dev/newReleaseBranch.py
+++ b/scripts/dev/newReleaseBranch.py
@@ -383,7 +383,8 @@ def update_project(project, version, branch_type):
         set_ags_test_versions(project, version)
     elif project == COMMUNITY_PACKAGING:
         update_ci_yaml(YAML_DICT.get(project), project, version, next_dev_ver)
-        update_acs_comm_pck_dependencies(branch_type, project)
+        update_ci_yaml(YAML_DICT.get(project), project, version, next_dev_ver) if branch_type == SERVICE_PACK else update_ci_yaml(
+            YAML_DICT.get(project), project, version + "-A1", next_dev_ver + "-A1-SNAPSHOT")
 
 
 def log_progress(project, message):
--- a/scripts/dev/newReleaseBranch.py
+++ b/scripts/dev/newReleaseBranch.py
@@ -124,6 +124,7 @@ PROJECTS = [ACS_PACKAGING, ENTERPRISE_SHARE, ENTERPRISE_REPO, COMMUNITY_REPO, CO
 # read command line arguments
 parser = argparse.ArgumentParser(description="Create git branches after ACS release.")
 parser.add_argument('-r', '--release', metavar='x.y.z', help='release version (x.y.z format)')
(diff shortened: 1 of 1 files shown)


scripts/dev/newReleaseBranch.py
```diff
@@ -0,0 +1,549 @@
+#######################################
+# This script creates HotFix branches for repo, share and packaging projects, and updates the master branches ready for the next SP/major release.
+# To use this script you need to install lxml:
+# 'pip install lxml'
+# Run the script without passing any arguments or with -h/--help argument to get usage information.
+# See below script behaviour explained.
+#######################################
+# Create a HotFix branches for the released version (for X.Y.Z release it will be X.Y.N eg., create 23.2.N for 23.2.0 release)
+# 1. acs-packaging:
+# - set RELEASE_VERSION to X.Y.Z+1, DEVELOPMENT_VERSION to X.Y.Z+1-SNAPSHOT in master_release.yml
```
tpage-alfresco, line 10: I think `Z` will always be `0` when calling this? I guess the DEVELOPMENT_VERSION will only be used when making the HF release and so should be higher than the RELEASE_VERSION.
```suggestion
# - set RELEASE_VERSION to X.Y.1, DEVELOPMENT_VERSION to X.Y.2-SNAPSHOT in master_release.yml
```

scripts/dev/newReleaseBranch.py
```diff
@@ -0,0 +1,549 @@
...
+# - set POM versions to X.Y.Z+1-SNAPSHOT
```
tpage-alfresco, line 11:
```suggestion
# - set POM versions to X.Y.1-SNAPSHOT
```

scripts/dev/newReleaseBranch.py
```diff
@@ -0,0 +1,549 @@
+#######################################
+# This script creates HotFix branches for repo, share and packaging projects, and updates the master branches ready for the next SP/major release.
+# To use this script you need to install lxml:
+# 'pip install lxml'
```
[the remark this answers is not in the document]
mpichura: Good catch - it isn't. I was trying out many xml parsing tools and ended up with `xml`, so that `lxml` install comment isn't necessary.

scripts/dev/newReleaseBranch.py
mpichura: Good point about `Z`.
However, script sets `DEVELOPMENT_VERSION` to 'Z+1' (which effectively is `1` instead of suggested `Z+2` = `2`).
At first I was setting that to `2` but after having looked at what happened historically I was confused and started setting that to `1`.
See following commits and let me know what you think:
https://github.com/Alfresco/acs-packaging/commit/b60e365d16018ae2b123c5bf015f337a80e5f3e9 (SP branch, dev version # same as release version #)
https://github.com/Alfresco/acs-packaging/commit/739bf7fd2d22efb6571f6c507728448ce0d37dc5 (HF branch, dev version # greater by `1` from release version #)
https://github.com/Alfresco/acs-packaging/commit/fdf4f38aa28c031c481ec949ef81d9474176f2f0 (master branch, dev version # same as release version #)
I'm not sure if it was historically correct, so  let me know what you think should be done here.

scripts/dev/newReleaseBranch.py
mpichura: Good point - will change that.

scripts/dev/newReleaseBranch.py
tpage-alfresco: So summarising those three commits - we go from the major release `23.1.0` to:
* `23.1.0.1`/`23.1.0.2-SNAPSHOT` on the HF branch
* `23.1.1-A1`/`23.1.1-SNAPSHOT` on the SP branch
* `23.2.0-A1`/`23.2.0-SNAPSHOT` on master.

I was instead expecting:
* `23.1.1`/`23.1.2-SNAPSHOT` on the HF branch
* `23.2.0-A1`/`23.2.0-SNAPSHOT` on the SP branch
* `24.1.0-A1`/`24.1.0-SNAPSHOT` on master.

Or a two branch solution (since we don't expect to release `2024.1.0` until after `2023.3.0` iirc):
* `23.1.1`/`23.1.2-SNAPSHOT` on the HF branch
* `23.2.0-A1`/`23.2.0-SNAPSHOT` on master.

I didn't think we supported four part version numbers in the new versioning scheme?

scripts/dev/newReleaseBranch.py
```diff
@@ -0,0 +1,549 @@
...
+# - set acs.version.label to .1
+# 4. community-share
+# - set scm-tag in main POM to HEAD
+# - set ACS version properties in main POM to X.Y.Z+1
+# - set main POM versions to  X.Y.Z+1.1-SNAPSHOT
+# 5. community-repo:
+# - set ACS version properties in main POM to X.Y.Z+1
+# - set POM versions to  X.Y.Z+1.1-SNAPSHOT
+# - set scm-tag in main POM to HEAD
+# - increment schema by 1 in repository.properties?
```
tpage-alfresco, line 29: I think this is unnecessary for the HF release branch. If someone makes a patch for a HF then they will increment the schema for it.

scripts/dev/newReleaseBranch.py
```diff
@@ -0,0 +1,549 @@
...
+# - set version.revision to Z+1 in version.properties (test resources)
+# 6. acs-community-packaging
+# - set RELEASE_VERSION to X.Y.Z+1, DEVELOPMENT_VERSION to X.Y.Z+1-SNAPSHOT in ci.yml
+# - set POM versions to X.Y.Z+1-SNAPSHOT?
```
tpage-alfresco, line 33: We only do HF releases for Enterprise, so I think the last digit will always be `0` for acs-community-packaging.

scripts/dev/newReleaseBranch.py
mpichura: Those 3 commits is what we historically used to do after a release.
At the moment new script does almost exactly what you described in "two branch solution":
- `23.1.1`/`23.1.1-SNAPSHOT` on the HF branch (easy to change to `23.1.2-SNAPSHOT`)
- `23.2.0-A1`/`23.2.0-A1-SNAPSHOT` on master (easy to change to `23.2.0-A1-SNAPSHOT` or even possibly better to `23.2.0-A2-SNAPSHOT` ?)

I think that, with time, we should add optional SP branch creation when next version (passed as `-d` argument to the script) is a major version bump. That way new SP branch would have same values as master for above examples and master would get whatever was passed in the script `-d` argument.
Does that make more sense?

scripts/dev/newReleaseBranch.py
mpichura: I didn't know how that was done practically. Historically we didn't bump that automatically when creating a branch. Should be easy to change, however, it is not harming us.

scripts/dev/newReleaseBranch.py
mpichura: I didn't know that either (HF for ent only). If that is the case, then we could possibly even abandon HF branch creation. WDYT?

scripts/dev/newReleaseBranch.py
tpage-alfresco: That sounds really good to me.

On the topic of SNAPSHOTs then I think they are a "snapshot" of what the next release will be and so `23.1.2-SNAPSHOT` is more accurate than `23.1.1-SNAPSHOT`. You're probably right about `23.2.0-A2-SNAPSHOT` - the weekly release will soon overwrite whatever we put, but it's overwhelmingly likely to be an A release next.

scripts/dev/newReleaseBranch.py
tpage-alfresco: I think it should be bumped for SP and LTS branches, but not HFs. You're right that it doesn't do any harm though.

scripts/dev/newReleaseBranch.py
tpage-alfresco: I don't think we should need HF branches in acs-community-packaging, but I just checked and they all seem to be there. Hardly any HF releases are there though. I suggest we get rid of them and add them back if we hit a problem.

scripts/dev/newReleaseBranch.py
```diff
@@ -0,0 +1,604 @@
...
+# 5. community-repo:
+# - set ACS version properties in main POM to X.Y.1
+# - set POM versions to  X.Y.1.1-SNAPSHOT
+# - set scm-tag in main POM to HEAD
+# - increment schema by 1 in repository.properties
+# - set version.revision to 1 in version.properties (test resources)
+# 6. acs-community-packaging
+# - not created as we do not release hot fixes for community version
+#######################################
+# Create a HotFix branches for the released version (for X.Y.Z release it will be release/X.Y+1.0 eg., create release/23.3.0 for 23.2.0 release)
```
mpichura, line 35:
```suggestion
# Create a Service Pack branches for the released version (for X.Y.Z release it will be release/X.Y+1.0 eg., create release/23.3.0 for 23.2.0 release)
```

scripts/dev/newReleaseBranch.py
```diff
@@ -0,0 +1,604 @@
...
+# 1. acs-packaging:
+# - set RELEASE_VERSION to X.Y+1.0-A1, DEVELOPMENT_VERSION to X.Y+1.0-A2-SNAPSHOT in master_release.yml
+# - set POM versions to X.Y+1.0-SNAPSHOT
```
mpichura, line 38: Should this be?
```suggestion
# - set POM versions to X.Y+1.0-A1-SNAPSHOT
```

scripts/dev/newReleaseBranch.py
```diff
@@ -0,0 +1,604 @@
...
+# - set main POM versions to  X.Y+1.0.1-SNAPSHOT
+# 5. community-repo:
+# - set ACS version properties in main POM to X.Y+1.0
+# - set POM versions to  X.Y+1.0.1-SNAPSHOT
+# - set scm-tag in main POM to HEAD
+# - increment schema by 100 in repository.properties
+# - set version.revision to Z+1 in version.properties (test resources)
+# 6. acs-community-packaging
+# - set RELEASE_VERSION to X.Y+1.0-A1, DEVELOPMENT_VERSION to X.Y+1.0-A2-SNAPSHOT in ci.yml
+# - set POM versions to X.Y+1.0-SNAPSHOT
```
mpichura, line 60: Should this be?
```suggestion
# - set POM versions to X.Y+1.0-A1-SNAPSHOT
```

scripts/dev/newReleaseBranch.py
```diff
@@ -0,0 +1,604 @@
...
+# - set scm-tag in main POM to HEAD
+# - set comm-repo dependency in main POM to X.Y+1.0.1
+# - set comm-share dependency in main POM to X.Y+1.0.1
+#######################################
+# Update master branch for the next SP/major release
+# 1. acs-packaging:
+# - set RELEASE_VERSION to <next_development_version>-A1 passed as script argument or X.Y+1.0-A1 (if <next_development_version> not passed),
+#   DEVELOPMENT_VERSION to <next_development_version>-A2-SNAPSHOT or X.Y+1.0-A2-SNAPSHOT (if <next_development_version> not passed) in master_release.yml
+# - set POM versions to <next_development_version>-SNAPSHOT or X.Y+1.0-SNAPSHOT (if <next_development_version> not passed)
```
mpichura, line 69: Should this be?
```suggestion
# - set POM versions to <next_development_version>-A1-SNAPSHOT or X.Y+1.0-A1-SNAPSHOT (if <next_development_version> not passed)
```

scripts/dev/newReleaseBranch.py
```diff
@@ -0,0 +1,604 @@
...
+# - set scm-tag in main POM to HEAD
+# - set ACS version properties in main POM to <next_development_version> or X.Y+1.0 (if <next_development_version> not passed)
+# - set POM versions to <next_development_version>.1-SNAPSHOT or X.Y+1.0.1-SNAPSHOT (if <next_development_version> not passed)
+# 5. community-repo:
+# - set ACS version properties in main POM to <next_development_version> or X.Y+1.0 (if <next_development_version> not passed)
+# - set POM versions to <next_development_version>.1-SNAPSHOT or X.Y+1.0.1-SNAPSHOT (if <next_development_version> not passed)
+# - set scm-tag in main POM to HEAD
+# - increment schema by 100 (when next development minor version bumped) or 1000 (when next development version major bumped) in repository.properties
+# - set version.major/version.minor/version.revision to <next_development_version> or X.Y+1.0 (if <next_development_version> not passed) in version.properties (test resources)
+# 6. acs-community-packaging
```
mpichura, line 89: This part also performs version update in POM files (will include that in the doc comment)

mpichura (commented):
I've left some comments for the topics I am still not sure about (or forgot to change)

scripts/dev/newReleaseBranch.py
tpage-alfresco, line 35: I think maybe:
```suggestion
# Create Service Pack branches for the released version (for X.Y.0 release it will be release/X.Y eg. for 23.2.0 release create release/23.2)
```

tpage-alfresco: Actually taking a step back. I think the simplest approach is:
```
HF Release (e.g. 23.1.1) -> No new branches needed
                            Update existing branch ready for 23.1.2
SP Release (e.g. 23.2.0) -> New HF branch needed (release/23.2)
                            Update existing branch ready for 23.3.0
LTS Release (e.g. 24.1.0) -> New SP branch needed (release/24.N)
                             New HF branch needed (release/24.1)
                             Update existing branch ready for 25.1.0
```

Alternatively if we want to use `master` for SPs as long as possible:
```
HF Release (e.g. 23.1.1) -> No new branches needed
                            Update existing branch ready for 23.1.2
SP Release (e.g. 23.2.0) -> New HF branch needed (release/23.2)
                            Update existing branch ready for 23.3.0
LTS Release (e.g. 24.1.0 following on from 23.3.0) -> New HF branch needed (release/24.1)
                                                      New SP branch needed (release/23.N) (created from a point before LTS release)
                                                      Master updated ready for 24.2.0
```
I think the second workflow is what we're aiming for, but this will require us to create a SP branch after the fact. I'm not even sure what should be in this SP branch. IMO the simplest approach is the first one, which would mean we very rarely work against `master` (or alternatively we cherry-pick almost all changes back, which would be really painful).

mpichura (commented):
I think I applied all suggested changes.
Also created a [Confluence page](https://alfresco.atlassian.net/wiki/spaces/TECH/pages/1588363298/Handling+branches+for+ACS+release+after+switching+to+new+versioning+schema) describing the process.

scripts/dev/newReleaseBranch.py
mpichura: Not bumping that anymore.

scripts/dev/newReleaseBranch.py
mpichura: Excluded comm-packaging from HF branch creation.

scripts/dev/newReleaseBranch.py
mpichura: Changed

scripts/dev/newReleaseBranch.py
```diff
@@ -556,24 +557,28 @@ def commit_all_and_push(project, message):
 
 def calculate_branch(branch_type, release_version=release_version, use_test_branches=args.test_branches):
     """Calculate the branch name
-    >>> calculate_branch("master", release_version="24.1.0")
-    'release/24.1'
+    >>> calculate_branch("master", release_version="23.2.0")
+    'release/23.2'
```
tpage-alfresco, line 561: Maybe this should just return `master`? It's confusing that it returns the same as `"hotfix"`, but different from `"service-pack"`.

scripts/dev/newReleaseBranch.py
```diff
@@ -556,24 +557,28 @@ def commit_all_and_push(project, message):
...
     >>> calculate_branch("hotfix", release_version="24.1.0")
-    'release/24.1.N'
+    'release/24.1'
     >>> calculate_branch("hotfix", release_version="24.1.0", use_test_branches=True)
-    'test/release/24.1.N'
+    'test/release/24.1'
+    >>> calculate_branch("release", release_version="23.2.0")
+    'release/stabilization/23.2'
+    >>> calculate_branch("release", release_version="23.2.0", use_test_branches=True)
+    'test/release/stabilization/23.2'
```
tpage-alfresco, line 573: Nice additional feature!

tpage-alfresco (approved):
LGTM! Nice work and looking forward to trying it out next week :rocket:

PR #2674, opened 2024-01-30
Testing of Sync Service multi Arch changes

jainakshat2:
please check and approve


abhishekkumar31 (approved):
We are merging the code to test multi arch changes. Will revert if SS tests continues to fail.

PR #2696, opened 2024-03-01
ACS-6764 Porting changes to master from stabilization

mpichura:
Porting needed changes from `release/stabilization/23.2` branch


pom.xml
```diff
@@ -19,11 +19,11 @@
...
 
-        <alfresco.s3connector.version>6.1.0-A1</alfresco.s3connector.version>
-        <alfresco.azure-connector.version>5.0.0-A1</alfresco.azure-connector.version>
-        <alfresco.salesforce-connector.version>3.0.0</alfresco.salesforce-connector.version>
+        <alfresco.s3connector.version>6.1.0</alfresco.s3connector.version>
+        <alfresco.azure-connector.version>5.0.0</alfresco.azure-connector.version>
+        <alfresco.salesforce-connector.version>3.1.0</alfresco.salesforce-connector.version>
         <alfresco.outlook.version>2.9.0</alfresco.outlook.version>
-        <alfresco.desktop-sync.version>5.0.0-A14</alfresco.desktop-sync.version>
+        <alfresco.desktop-sync.version>4.0.0</alfresco.desktop-sync.version>
```
tpage-alfresco, line 26: I don't think we want this change on master. @abhishekkumar31 ?

mpichura: I spoke to him about that and we agreed to port it to master, correct  @abhishekkumar31?

abhishekkumar31: Yes @tpage-alfresco , we can keep 4.0.0. So, we released SS 5.0.0-M3 in Eridanus, so either we can keep 4.0.0 or 5.0.0-M3. We have already tested the 4.0.0 with ACS master so it won't break anything.
Even if we use alpha or Milestone, we will have to change to 4.0.0 again if we release ACS as GA.

abhishekkumar31: In case there is a Milestone release planned for ACS in Fornax then keeping 5.0.0-M3 makes more sense.

mpichura: @abhishekkumar31 I will merge this PR as is and let you/your team to handle this dependency in the future (no matter it is close or distant future).

krdabrowski (approved):
LGTM

tpage-alfresco (approved):

MateuszHY (approved):

PR #2744, opened 2024-04-16
ACS-7598 Backward Compatible ATS/AIS

pzhyland:


SaraAspery (approved):

pom.xml
```diff
@@ -20,12 +20,12 @@
         <alfresco.rm-enterprise-rest-api-explorer.version>3.0.2</alfresco.rm-enterprise-rest-api-explorer.version>
 
         <alfresco.s3connector.version>5.1.0</alfresco.s3connector.version>
-        <alfresco.azure-connector.version>3.1.0</alfresco.azure-connector.version>
+        <alfresco.azure-connector.version>3.1.1-A2</alfresco.azure-connector.version>
```
mstrankowski, line 23: can't that be 3.1.2?

pzhyland: Unfortunately it can't be used right now. See https://hyland.slack.com/archives/C06NV7GKH08/p1712837889871639?thread_ts=1712836702.283169&cid=C06NV7GKH08

If we progress with the fixes we can reach the point it could be used but we are far from that :)

mstrankowski: Ah right, I messed up, the released version was 3.2.1 and it's not compatible out of the box, right, thanks.

PR #2772, opened 2024-05-09
ACS-7893 Bump S3 Connector to 6.1.1-A1 & Azure Connector to 5.0.1-A1

dsibilio:
Bumping S3 connector to 6.1.1-A1 for testing purposes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -515,6 +515,9 @@ jobs:
       - name: "Print output after failure"
         if: ${{ always() && steps.tests.outcome == 'failure' }}
         run: ${TAS_SCRIPTS}/output_logs_for_failures.sh "tests/tas-all-amps"
+      - name: "Dump all Docker containers logs"
+        uses: Alfresco/alfresco-build-tools/.github/actions/docker-dump-containers-logs@v1.41.0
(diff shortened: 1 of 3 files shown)


mikolajbrzezinski (approved):

dsibilio:
Needs work. Either needs a new Azure Connector `5.0.1-A1` with `cloud-connector-base:4.0-A4` as parent, or needs a new S3 Connector `6.1.1-A2` with `cloud-connector-base` reverted to `4.0-A3`.

dsibilio:
@tiagosalvado10 I had to perform the following change to make this PR pass: https://github.com/Alfresco/acs-packaging/pull/2772/files#diff-e8ddc83befb2ae12b4f76ca486ec78680313274c2d40c0063af22f329c208b3dR29, I believe this may be solely because we're manually creating our own Azure Connector-based beans here which is not the norm. Please correct me if I'm wrong.

In case I'm wrong (aka: changes are actually needed on the customer-side to add a default value for the new property manually) then that would obviously be an issue as we're releasing this as a simple hotfix version and don't expect any breaking changes.

tiagosalvado10 (approved):
LGTM

damianujma (approved):

mikolajbrzezinski (approved):

PR #2778, opened 2024-05-10
Bump Salesforce Connector to 2.4.2 & Outlook Connector to 2.9.3

dsibilio:
As per title.


kavitshah-gl (approved):
Yes this is fine. Thanks for updating it.

PR #2877, opened 2024-07-18
ACS-8147: Add hxi-connector to acs-packaging

krdabrowski:
* ACS-8147: Add hxi-connector to acs-packaging


tpage-alfresco (approved):
LGTM

dsibilio (approved):
If any of the new configuration properties are mandatory for the ACS deployment to succeed once the hxinsight-extension is installed, I'd recommend having the Terraform PR ready before envs start to fail due to the `pipeline-all-amps` image changes. This also applies to the other envs that could be impacted (e.g.: acs74n, acs73n).

Hopefully there are none!

krdabrowski:
@dsibilio fortunately there are none new required properties, but if we would like to use the new feature in some tests here, then two new configuration properties should be added

PR #2946, opened 2024-10-10
I have Updated the Salesforce Version

varapathijanakiram:
I have Upgrade the salesforce connector version 4.0.0-A1


rrajoria (approved):

MohinishSah (approved):

jottley:
Going forward please reference the related JIRAs in your commit message. I understand that in some cases there may not be a JIRA. In this case the version moving from to the version moving to would be useful when reading back through the commit history.

PR #2982, opened 2024-11-13
[fix/APPS-2894_NodeSizeDetails_AutomationTest] Adding .txt file, for community AutomationTest Cases

mohit-singh4:
https://hyland.atlassian.net/browse/APPS-2894


kavitshah-gl (approved):
Taking reference of alfresco repository, I am approving since we are placing the test file at the same location.

PR #3004, opened 2024-12-13
[ACS-9069] Fix failing tests in acs-packaging related to durable subscription in live indexing

cezary-witkowski:


cezary-witkowski:
See https://github.com/Alfresco/acs-packaging/pull/3001

PR #3043, opened 2025-02-10
[ACS-9202] Please [release] and [publish] ACS 23.5.0-A1

cezary-witkowski:


kmagdziarz (changes requested):
This branch is on-hold until new SP will be needed

